PR #347, opened 2014-10-18
Magnifying glass

liach:
Add sth. for mag to see the dual head signal block and the signal block. There may be a lot of bugs.


liach:
Well I believe that I code much better now. :trollface: The old code looks pretty horrible.

CovertJaguar:
Haha.

PR #360, opened 2014-10-25
Add custom comparator input override to iron tanks

wshadow:
Using this code, comparators can be used on tank valves to read how much liquid is in the tank.

Work in progress pull request. The current implementation requires the user to provide blockupdates to the comparator - without those the comparator doesn't update its value. I'd like to do the blockupdates only when necessary (i.e. the signalstrength will change) and only on the blocks adjacent to the tank valves instead of all blocks adjacent to the tank. I'm looking for ideas how (i.e. in which classes/methods) I could implement this. I'll update this pull request when I have figured out how to do that.


choncho:
@wshadow You can do it!

PR #467, opened 2015-01-17
Reworked logic in TrackNextGenLocking.java

ssotangkur:
These changes allow the track to become unpowered immediately if a train hasn't recently crossed over it.
It also fixes an issue where upon restoring from NBT, the train gets held by a locking track that was in the middle of releasing it but was no longer powered by redstone. 
This seems to be caused by train instances that change a few ticks after loading. For instance, when initially calling LinkageManager.getTrain(cart) you will get one instance
of Train (id=1), but on another tick you would get Train (id=2) even though you used the same cart instance in both calls. Once the game stabilizes, the Train instances also
stabilize so we don't see the issue normally. The fix was to only store cart references across ticks (not trains) then when we need to compare trains, simply call LinkageManager.getTrain()
on each cart which guarantees that the train returned will be the same. 
The rest of the changes were to eliminate as many timers/delays as possible and rely on the state of the world as much as possible.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/tracks/TrackNextGenLocking.java
+++ b/src/main/java/mods/railcraft/common/blocks/tracks/TrackNextGenLocking.java
@@ -13,6 +13,7 @@ import java.io.DataOutputStream;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.util.UUID;
+
 import mods.railcraft.api.core.items.IToolCrowbar;
 import mods.railcraft.api.events.CartLockdownEvent;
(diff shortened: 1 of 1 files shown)


ssotangkur:
I found a bug. Wait before accepting these changes.
It seems that liquid loaders that aren't able to signal the track to activate. It might be because of the removal of the onNeighborBlockUpdate() method.

CovertJaguar:
I suspect its more likely your overriding of the releaseCart() method which the Loaders call.

ssotangkur:
Yup, that's exactly right. I've closed this request and created a new one with the fix.

CovertJaguar:
Replaced by #469

PR #469, opened 2015-01-18
Reworked logic in TrackNextGenLocking.java

ssotangkur:
Added capability to unpower a train locking track when there is no redstone signal and no train overhead.
Fixed issue on load from NBT where trains in the process of leaving a train locking track would get stuck.
Simplified the logic a little by eliminating as many countdown delays/timers as possible.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/tracks/TrackNextGenLocking.java
+++ b/src/main/java/mods/railcraft/common/blocks/tracks/TrackNextGenLocking.java
@@ -164,7 +164,7 @@ public class TrackNextGenLocking extends TrackBaseRailcraft implements ITrackLoc
         	}        	
         	        	
             if (currentCart != null && currentCart.isDead) {
-                releaseCart();
+                releaseCurrentCart();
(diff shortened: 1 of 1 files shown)


src/main/java/mods/railcraft/common/blocks/tracks/TrackNextGenLocking.java
```diff
@@ -324,12 +390,25 @@ public void readFromNBT(NBTTagCompound data) {
...
+        	trainDelay = data.getInteger("trainDelay");        
+
+        prevCartUUID = readUUID("prevCart", data);  
+        currentCartUUID = readUUID("currentCart", data);
+        
+        uuid = readUUID("uuid", data);
+        
+        justLoaded = true; // This signals updateEntity() to dereference the cart UUID's we read in here
+        
+        System.out.println("Read from NBT");
```
CovertJaguar: *cough* You might want to remove this ;)

FYI, you don't need to create a new PR, just make a commit to your branch.

PR #474, opened 2015-01-21
Fixed issue with carts moving in impossible directions on switch tracks

ssotangkur:
I used a trick where we cache the carts that are just about to spring or lock the track. Then in the subclasses' getBasicRailMetadata() method we can test against this cache to see whether the cart attempting to get the possible track direction should use the sprung or lock direction. This ensures that the cart always gets the correct meta, even if visually the track doesn't reflect this.


ssotangkur:
Okay, it looks like I finally got all the whitespace to match. This PR should be good to go (in that respect at least).

src/main/java/mods/railcraft/common/blocks/tracks/TrackSwitchBase.java
```diff
@@ -27,6 +35,13 @@
     protected boolean switched;
     private byte sprung;
     private byte locked;
+    protected Set<EntityMinecart> lockingCarts = new HashSet<EntityMinecart>();
+    protected Set<EntityMinecart> springingCarts = new HashSet<EntityMinecart>();
```
CovertJaguar: Why not just cache the UUID? Wouldn't that negate the need for finding the entity?

src/main/java/mods/railcraft/common/blocks/tracks/TrackSwitchBase.java
```diff
@@ -128,6 +143,56 @@ public void onNeighborBlockChange(Block block) {
         }
         super.onNeighborBlockChange(block);
     }
+    
+    /**
+     * A utility method for writing out UUID's to NBT 
+     */
+    private void writeUUIDToNBT(String key, UUID uuid, NBTTagCompound data) {
```
CovertJaguar: I probably should have pointed this out before, but....
https://github.com/CovertJaguar/Railcraft/blob/master/src/main/java/mods/railcraft/common/util/misc/MiscTools.java#L69

ssotangkur: Ah, sweet. Yup, I had been looking for classes like NBTTools or NBTUtils. Never thought to look there.

CovertJaguar:
I wonder if a fixed length FIFO Queue or Deck might be ideal here. Unfortunately as far as I know, neither Java or Quava provides such a collection so you've probably have to write it yourself.

ssotangkur:
The LinkedList class (http://docs.oracle.com/javase/7/docs/api/java/util/LinkedList.html) works as a FIFO Queue if you use the add()/remove() methods as the queue()/dequeue() methods. Although, I'm not sure if that makes as much sense as a replacement for the Set. My concern is the off chance that getMinecartsAt() may return more than one cart and the order may not be correct.

On another note, I'm trying to add a method to CartTools to return cart UUIDs instead of just carts, but I can't seem to locate them in my git. I'm guessing its a git submodule. Do you prefer that I create my own method here or update the submodule?

ssotangkur:
I've added the change w/ CartUtils. Let me know if there's anything else you want me to change.

src/main/java/mods/railcraft/common/blocks/tracks/TrackSwitchBase.java
```diff
@@ -184,24 +217,60 @@ public void updateEntity() {
...
+        }
+
         if (locked > 0)
             locked--;
-        if (shouldLockSwitch())
-            locked = SPRING_DURATION;
-
-        boolean springState = sprung == 0;
         if (sprung > 0)
             sprung--;
```
CovertJaguar: Technically, these counter decrements should probably be before the cart searches, else you are shortening the duration by a tick.

CovertJaguar:
Man its been so long since I looked at this code, I'd forgotten how convoluted the orientation logic was.

src/main/java/mods/railcraft/common/blocks/tracks/TrackSwitch.java
```diff
@@ -44,7 +46,7 @@ public IIcon getIcon() {
     @Override
     public int getBasicRailMetadata(EntityMinecart cart) {
         int meta = tileEntity.getBlockMetadata();
-        if (cart != null && isSwitched()) {
+        if (cart != null && (isSwitched() || springingCarts.contains(cart.getPersistentID()))) {
```
CovertJaguar: Wouldn't this "derail" a cart from the "locked" direction if the track was switched? Probably need to add a check against the locked set when isSwitched() is true.

CovertJaguar: Also...you may need more granularity than isSwitched(), since isSwitched() also takes the cosmetic counters into account. You probably want to use the switched variable directly.

CovertJaguar: Hmm...however, TileSwitchBase I think needs the cosmetic result from isSwitched() in all cases. Maybe don't change isSwitched(), just use the switched variable.

https://github.com/CovertJaguar/Railcraft/blob/master/src/main/java/mods/railcraft/common/blocks/signals/TileSwitchBase.java#L116
https://github.com/CovertJaguar/Railcraft/blob/master/src/main/java/mods/railcraft/common/blocks/signals/TileSwitchBase.java#L190

CovertJaguar:
And now that I've probably thoroughly confused both myself and you, I'll leave you to sort it out. ;)

ssotangkur:
LOL, yes. 
I made the wrong assumption about which tracks are the locked track. I just sorted that out now.
Essentially, there are 3 types of tracks adjacent to any switch track:
1) should spring - needs to spring the track when it arrives
2) should lock - needs to unspring the track when it arrrives
3) decision - can go either way based on the state of "sprung/locked/switched"

I need to adjust the logic so that "decision" carts rely on the state variables

I'm planning to create a new protected method isSwitched(cart) which is really only called by the subclases. It will take the place of (isSwitched() || springingCarts.contains(cart.getPersistentID()) but have even more logic in it.

CovertJaguar:
I'm not sure that the "decision' approach necessary cares about the state of sprung or locked. 

Though I admit attempting to reverse the direction of a Train parked on a switch will undoubtedly cause undefined behavior, but that's true with real life Spring Switches too.

ssotangkur:
I think decision carts still need to respect the sprung/lock timers in case they are part of trains.

ssotangkur:
Unless you want to make it so that flipping a switch in the middle of the train will direct part of the train in the other direction. (I can make it so if that's what you want).

CovertJaguar:
Flipping the switch should affect only carts approaching from the "decision" direction, it should of course have some safeguard to prevent that from happening in the middle of a train. However, the lock and sprung state are not that safeguard as they should apply to carts coming from the lock and sprung direction, right?

ssotangkur:
I see your point. 
In a way, you could say I'm overloading the usage of sprung/locked to also account for "decision" carts when they travel on switched or unswitched track to provide that safeguard to prevent mid-train switching. This seems to work well in my testing when a train stops over a switch track and reverses direction. In this scenario, the decision cart crosses over and becomes a spring or lock cart, then during the reverse direction the same sprung/locked variable gets updated as it goes back over the switch track (hopefully this makes sense).

ssotangkur:
I've added "decision" cart tracking which allows me to keep the other 2 sets up to date by allowing me to remove the carts from the other sets if they appear here.

I've also added train checking and consolidated logic into the isSwitched(Cart) method.

ssotangkur:
Bounce. Hoping this doesn't get lost in the shuffle.

CovertJaguar:
No, sorry. I've just been really busy the last few days and didn't want to touch it until I had time to give it the proper attention. I should have time to process it next week.

ssotangkur:
No rush. Please take your time. Just wanted to let you know that I've updated it.

CovertJaguar:
Looks good!

Issue #475, opened 2015-01-22
[Suggestion] Add burntime bonus to blocks of CoalCoke


wshadow:
In vanilla minecraft Coal blocks have a higher BurnTime then 9 individual pieces of Coal combined (smelts 10 times 8 Items vs. 9 times 8 Items in a furnace). I think it wold be nice to have a similar bonus with blocks of Coal Coke.

Quote from Gamepedia: "One block of coal smelts 80 items, which is more than the smelting time of the 9 coal items used to craft it."

CovertJaguar:
What is the fuel value of a Block of Coal?

wshadow:
I don't know the exact internal numbers, but a Block of Coal has 10 times the burn time compared to a single piece of coal.

Ps.: just noticed that that part of the original text got lost due to formating fails :(

liach:
Coal Block: 80 items=32000 fuel value
Coke Block: 144 items=57600 fuel value
(in vanilla furnace)
That will make Coke more useful in Bores or Steam Boilers.

wshadow:
Yes, a Coke Block is better then a Coal Block. But according to my experience and mods.railcraft.common.plugins.forge.FuelPlugin the burntime of a Coke Block is equal to 9 pieces of Coke.

liach:
The purpose is to change the fuel value of coke block to 64000.

liach:
And to change coke block into flammable block.

CovertJaguar:
I hope its not 64000, because that massively overflows the short in the GUI burn progress packet.

liach:
Oh I thought wrong, 57600 is the value in 1.6, which can cook a steel with a coal coke. Since one version in series 8 the heat reduced, it should be 32000 now.

wshadow:
My pull request fixes the original issue. I didn't change anything regarding the suggestion from @liach: "change coke block into flammable block"

PR #476, opened 2015-01-22
Update Coke Block fuel value

liach:
Add flammability and increase the fuel value of the Coke Block.
I don't think the creosote block should have so high a resistance.
#475

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/aesthetics/cube/EnumCube.java
+++ b/src/main/java/mods/railcraft/common/blocks/aesthetics/cube/EnumCube.java
@@ -22,7 +22,7 @@ import mods.railcraft.common.modules.ModuleManager.Module;
  */
 public enum EnumCube {
 
-    COKE_BLOCK(Module.FACTORY, "coke", new SimpleCube(), 2f, 10f),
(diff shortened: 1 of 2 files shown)


src/main/java/mods/railcraft/common/blocks/aesthetics/cube/EnumCube.java
```diff
@@ -22,7 +22,7 @@
  */
 public enum EnumCube {
 
-    COKE_BLOCK(Module.FACTORY, "coke", new SimpleCube(), 2f, 10f),
+    COKE_BLOCK(Module.FACTORY, "coke", new FlammableCube(5, 300), 2f, 10f),
```
CovertJaguar: 300 fire spread is a bit extreme. The Coal Block is only a 5.

liach: Will fix in next commit. To prevent too many commits you could commit what I authored, and that feels better.

CovertJaguar: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

CovertJaguar:
So....what? Are you retracting your PR?

PR #484, opened 2015-02-01
Added catenaries and pantographs

### LINKED ISSUE #480: [Suggestion] Catenaries and pantographs for electric locomotives
One way for electric locomotives to get electricity is through overhead lines. They should be cheaper to make than electric tracks, as such:

These wires do not actually store any energy, but rather just get it from other energy sources, like generators and wires. They still lose energy over distance, so it makes sense to shorten the length of the catenaries.

Note how the catenaries are rendered - they are rendered in the top part of the block so that placing them underneath under blocks for decoration, like metal posts, makes sense.

For the locomotive to actually collect energy, we'll need a pantograph.

Pantographs have durability (currently 20,000) and it decreases whenever the locomotive is moving and touching the wires, to simulate friction. If you want to destroy the pantograph even faster, just wedge it into some block and the durability will be falling by 100 every so often.

The electric locomotive now has a special slot for a pantograph:

Placing the pantograph will render the locomotive as such:

Admittedly, this is not the best design, but it's pretty understandable.

I think the main advantage over electric rails is that you can now run your train over high speed rails and still be able to collect energy from the overhead lines.

The disadvantage, though, is that these catenaries are just as lossy as regular wires, but they do not store any energy, so it's up to the player to implement the design to reduce energy loss. Plus, since the pantograph has durability, the player has to maintain the trains to make sure they don't run out of power in the middle of the tracks.

<hr>

As you can see, the code is pretty much ready. I would like to hear some opinions before submitting pull requests. Note that not only will this require a change in the code, but also in the API (two new interfaces are added, IElectricDistributor for the non-energy storing wires and IPantographProvider to be able to provide a pantograph) as well as a few lines in the localization files.

wooky:
Closes #480


liach:
Merged in dev?

wooky:
Unfortunately no. Last time we discussed about this, CovertJaguar said he wants to rework the energy system to be tickless before catenaries are added. I haven't followed on this so I don't know what's the status.

CovertJaguar:
Status == Busy Busy Busy. ;)

I'll get around to this eventually, I promise.

wooky:
I'm closing this because I have a plan to improve catenaries.

liach:
Oops. CJ has not shown up on github for nearly a month.

wooky:
Regardless, my plan is to make it more elegant and easy to set up catenaries. Hopefully won't take too long.

CovertJaguar:
Yeah that was the main thing holding me back. I didn't like they way that you have to place each block of "wire".

wooky:
I really want to be able to draw lines like in Immersive Engineering, but I'm pretty terrible at understanding how those lines are being drawn on the screen.

CovertJaguar:
@wooky see https://github.com/CovertJaguar/Railcraft/blob/master/src/main/java/mods/railcraft/client/render/RenderTESRSignals.java#L59

I'm using line drawing to draw the Signal Auras here.

PR #491, opened 2015-02-06
Update build.gradle

Vexatos:


Yopu:
What advantage does this provide over the current implementation?

CovertJaguar:
Supposedly it fixed issues with resources when running IntelliJ.

Yopu:
No objections here. Just curious.

Vexatos:
Using IDEA, FML does not recognize resources if they are in a different folder than the java classes. This PR makes IDEA compile both classes and resources into the same folder.

Issue #490, opened 2015-02-05
Switches don't always handle fast Trains near slopes correctly


CovertJaguar:
There seems to be some edge cases with fast trains and nearby slopes that don't work well with the new switch code.

Talk to @purpleposeidon (neptunepink) for more information.

```
[05 3:37 am] <+neptunepink> There was 1 flat track between the switch and the slope. Now there's two. It still doesn't work at the top speed.
[05 3:39 am] <+neptunepink> 4 track buffer. There was also a control track at the top that I removed. I can't move the hill back any further.
[05 3:49 am] <+neptunepink> Also it seems like the vehicles are doing the vanilla-style track glitch thing
[05 3:49 am] <+neptunepink> where they suddenly change orientation as if they'd entered from the other direction
[05 3:49 am] <@CovertJaguar> hmm....client server sync issues maybe
```

@ssotangkur *poke*

ssotangkur:
Damn, I should have tested slopes too.
On it.

ssotangkur:
Is the glitching occurring only when the switch is track is powered (trying to divert traffic) and you make the cart go through the opposite direction where the track has to automatically unswitch?

I'll ask for this in IRC too.

ssotangkur:
I've emailed NeptunePink. I've tried a couple setups with slopes and reinforced switch tracks with booster tracks and locos at full speed and I'm not able to create any serious problems. I did notice the UI glitching, which is a side-effect of the way we provide track metadata to carts individually.
I can't tell for sure, but I believe the clients ask for the track meta w/o passing in a cart in the arguments. This means that it can only get the correct "state" of track, after a cart actually sets the state by entering or leaving the track. My guess is that the client is preemptively asking for the track state, right before the cart enters the track. The client then doesn't ask for the track again until the cart leaves.

Part of my reason for thinking this is that I've also tried doing a client update on every tick just to see if that would resolve it, but it doesn't.

I think though, the glitching can be resolved if I test the UUID sets to see if they are empty to guess what state the track will be in.

CovertJaguar:
The client probably doesn't call the minecartOnTrack() function at all, in which case in has to rely on the old method of metadata swapping.

CovertJaguar:
[...]

PR #497, opened 2015-02-09
Fix for #490 [Switches don't always handle fast Trains near slopes correctly]

ssotangkur:
https://github.com/CovertJaguar/Railcraft/issues/490

There is an associated PR for the API change also here: https://github.com/CovertJaguar/Railcraft-API/pull/6

The fix involves changing how switch devices (levers/motors/routing) interact w/ the switch track. Instead of letting the switch modify the track's switch state directly, we let the track ask the switch when it needs to. This allows the track, which is already searching for carts, to submit the carts it finds to the switch logic, w/o the switch needing to search for the same carts on its own. 

The issue that was found, arose from a tick ordering where carts that would have been found on the same tick by the track and the switch, were dependent on the switch processing first. If the switch's updateEntity() ran first, it would be able to set the "switched" variable in the track and the track would be able to "spring". On the other hand, if the track ran first, the switched variable would be false and the track would become "locked" for the cart.

Please note, I'm not sure what to do about the submodule changes so you probably don't want to accept these changes quite yet.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/signals/TileSwitchRouting.java
+++ b/src/main/java/mods/railcraft/common/blocks/signals/TileSwitchRouting.java
@@ -8,11 +8,7 @@
  */
 package mods.railcraft.common.blocks.signals;
 
-import java.util.List;
-import mods.railcraft.api.tracks.ITrackInstance;
 import mods.railcraft.api.tracks.ITrackSwitch;
-import mods.railcraft.common.blocks.tracks.TileTrack;
-import mods.railcraft.common.carts.CartUtils;
 import mods.railcraft.common.gui.EnumGui;
 import mods.railcraft.common.gui.GuiHandler;
 import mods.railcraft.common.gui.buttons.MultiButtonController;
@@ -26,9 +22,6 @@ import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.inventory.IInventory;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
-import net.minecraft.tileentity.TileEntity;
-import net.minecraft.util.AxisAlignedBB;
-import net.minecraftforge.common.util.ForgeDirection;
 
 public class TileSwitchRouting extends TileSwitchSecured implements IRouter, IRoutingTile {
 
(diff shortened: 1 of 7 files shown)


src/main/java/mods/railcraft/common/blocks/tracks/TrackSwitchBase.java
```diff
@@ -102,33 +106,30 @@ protected boolean isSwitched(EntityMinecart cart) {
...
-    //        return super.blockActivated(player);
-    //    }
     @Override
     public void onBlockPlaced() {
         determineTrackMeta();
         determineMirror();
+
+        // Notify any neighboring switches that we exist so they know to register themselves with us
+        Block myBlock = getWorld().getBlock(getX(), getY(), getZ());
+        getWorld().notifyBlocksOfNeighborChange(getX(), getY(), getZ(), myBlock);
```
CovertJaguar: there should be a shortcut function for this on the super class notifyBlocksOfNeighborChange().

ssotangkur:
Ok, CJ, this is ready for review. I'll try to catch you on IRC to see if you have any questions.

src/main/java/mods/railcraft/common/blocks/signals/TileSwitchRouting.java
```diff
@@ -151,4 +114,13 @@ public IInventory getInventory() {
         return inv;
     }
 
+    @Override
+    public boolean shouldSwitch(ITrackSwitch switchTrack, EntityMinecart cart) {
+        RoutingLogic logic = getLogic();
+        if(logic != null)
```
CovertJaguar: You are missing the logic.isValid() check.

src/main/java/mods/railcraft/common/util/misc/MiscTools.java
```diff
@@ -83,6 +86,23 @@ public static UUID readUUID(NBTTagCompound data, String tag) {
         return null;
     }
 
+    public static void writeUUID(DataOutputStream data, UUID uuid) throws IOException {
```
CovertJaguar: These should probably be in DataTools in the network package rather than MiscTools.

CovertJaguar:
Would it be possibly to squash this?

```
git rebase -i HEAD~15
git push -f
```

ssotangkur:
Can't get rebasing to work. Seems to be causing a bigger mess as it can't do it without lots of conflicts.

ssotangkur:
This is ready for review. Rebase didn't work, had to use "merge --squash" instead and some "git reset" voodoo.

CovertJaguar:
Oh? You got it working? Nice. Congrats on leveling up your Git-Fu!

ssotangkur:
Thanks. Learned way too much git than I wanted to. I wished rebase could be a bit smarter when it knows you are squashing commits.

CovertJaguar:
Merged manually in 39746ede9da90b7d8c9df710117ef9e6207bcc8c

PR #498, opened 2015-02-09
Allow signal capacitors to remain on as long as there is a signal

wshadow:
- fix for CovertJaguar/Railcraft#121
- Capacitor starts countdown when the last input is turned off
- Uses the most restrictive aspect of all active inputs (checked every time the counter would decrement)
- redstone repeater input counts as GREEN aspect

### DIFF
--- a/src/main/java/mods/railcraft/client/gui/GuiBoxCapacitor.java
+++ b/src/main/java/mods/railcraft/client/gui/GuiBoxCapacitor.java
@@ -9,6 +9,7 @@
 package mods.railcraft.client.gui;
 
 import net.minecraft.client.gui.GuiButton;
+import mods.railcraft.client.gui.buttons.GuiMultiButton;
 import mods.railcraft.common.blocks.signals.TileBoxCapacitor;
 import mods.railcraft.common.plugins.forge.LocalizationPlugin;
 import mods.railcraft.common.util.misc.Game;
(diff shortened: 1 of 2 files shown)


wshadow:
Keep in mind that his changes the behavior of the signal capacitor and does not add a different box as someone suggested. So this may(?) break existing builds that depend on the old behavior of the capacitor box.

CovertJaguar:
I'd prefer a toggle button in the GUI.

wshadow:
Hm, toggle button. Sounds like a good idea. I'll try to figure out how to design a GUI when I get back home this evening.

wshadow:
Code is finished and works perfectly in my dev env. labels for the new button at CovertJaguar/Railcraft-Localization#278 . I'm not sure if/how the ToolTip works. I included the Strings, but I did not find out how to display them ingame.

KamranMackey:
@wshadow You might want to squash your commits before @CovertJaguar merges it. Just a tip. :)

wshadow:
@KamranMackey huh? Is there a policy for that, that I missed? If not, I'd think it's better not to throw away history?!

KamranMackey:
No, but it is what CovertJaguar suggests when a PR has a lot of commits in it.

CovertJaguar:
Its preferred, otherwise the main history gets cluttered and if I have to revert a PR, its an order of magnitude easier if the PR was a single commit.

wshadow:
Ok, it's your history; give me 10 minutes. Should be easy enough to do.

CovertJaguar:
Thanks!

PR #507, opened 2015-02-12
block == Blocks.air to world.isAirBlock in hidden blocks

warlordjones:
== Blocks.air is bad practice. Should help some issues I've been having with my TEs being replaced by TileHiddens too.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/hidden/TrailTicker.java
+++ b/src/main/java/mods/railcraft/common/blocks/hidden/TrailTicker.java
@@ -76,8 +76,7 @@ public class TrailTicker {
     }
 
     private boolean trySetMarker(World world, int x, int y, int z, EntityPlayer player) {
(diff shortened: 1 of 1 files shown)


Vexatos:
Does it appear anywhere else in the code?
Also, this is quite an important fix, actually.

warlordjones:
It appears a lot, but this instance seems to be one of the only ones that affects anything.

CovertJaguar:
This check is intentional, it's purpose is to prevent anything except vanilla air blocks from being replace by heat blocks. What reason is there to change this?

warlordjones:
Makes sense.

warlordjones:
There is no reason - hence why I closed the issue.

I was having some issues with RC heat tiles replacing mine, or at least I think they were. Although I fixed this causing a crash on my end, they may be worth investigating further.

Issue #489, opened 2015-02-04
[Feature Request] Special Anchor (chunkloader)


Terpo:
Hello,

is there any chance to add a special kind of chunkloader?
- anchor does not force load after server restart
- anchor activates if any player is nearby (not just the owner)
- anchor stays online until next restart or out of fuel

Now you may ask: Why would someone need something like this? 
I'm admin of a small private server. The server has enough power, but there is something really weird when it comes to chunk unloading. After a player left the server, some parts of the base stays loaded. I talked with ProfMobius about it because even a /chunkpurge command does not unload these chunks. (picture: http://blog.terpo.org/upload/mc/chunkpurge.png)

I'm logging the serveroverload on our homepage. The more bases are loaded, the higher is the overload. 
Two examples:
[13:38:19] Running 2115ms behind, skipping 42 tick(s) 
[16:52:28] Running 20883ms behind, skipping 417 tick(s) 

Here are two charts of a testserver. I did only join, teleport to a base and left the server.
http://blog.terpo.org/upload/mc/serverload.html
I also tried this on different PCs and the TPS is at max. 20ms. The whole server is stopping for this period of time. 

As you can see, adding chunkloaders stabilizes the server, but I don't want to load all bases with admin anchors.

Hopefully that is enough information for you :)

CovertJaguar:
So these chunks that staying loaded aren't currently being loaded by anything other that, for example: "hopper chunkloaders"? In other word, you aren't currently running chunkloaders, but the chunks stay loaded because of chunk-thrashing?

Terpo:
I can't figure out whats causing the chunks to stay loaded. There are no explicit chunkloaders in most of those bases. Yes, I want a higher TPS rate with chunkloaders when it helps to prevent massive server lag spikes. I did a video about it for EnderIO because I found out a cross mod problem that shows the problem: https://www.youtube.com/watch?v=U-b0sxdrjA0

Minecraft stops being fun when the server stops for 8 or more seconds every 30 seconds or so.

CovertJaguar:
Your argument is valid, your use case is specific and reasonable. I see no reason to deny the request. ;)

Now someone just needs to write some code.

Terpo:
Most of the code should already be there. Your anchors have the most whats needed.

Terpo:
I'm trying to add the anchor myself but my knowledge in modding is not that great. Adding another block to your alpha machines is not possible. So I put it into gamma for the moment. 

[...]

PR #520, opened 2015-02-24
- added community anchor

Terpo:
Added a new anchor which will activate if someone is in range. It is only checked if there are player online, so it may happens that other chunkloader will also activate this one.
The anchor will stay active until the next restart occurs. It has no recipe, does not use fuel and can be combined with a sentinel.


Terpo:
It would be too powerful and unbalanced compared to the normal world -and personal anchor. 

Also it should be used by an admin if a base won't unload correctly any more.

CovertJaguar:
I'll look at this for 9.6, the final version will probably not be quite the same.

Terpo:
Any idea how you want to change it?

wshadow:
I'm gonna leave a link to the original issue here #489 for future reference.

CovertJaguar:
Well I'd like to add a recipe and possibly change the name, and I haven't even looked at the texture yet.

Terpo:
Oh, ok. 
The texture is the same as the other anchors, but with blue corners.

bookerthegeek:
Village anchor?

PR #531, opened 2015-03-03
Fix #519

### LINKED ISSUE #519: Crowbar with Destruction turns Concrete block into Coke block
This was an fun one. I have a set of rails on a crushed obsidian path, with concrete blocks on either side of the obsidian. When removing the rails with a crowbar enchanted with Destruction, the surrounding concrete blocks break and turn into Blocks of Coal Coke.

This seems to only effect full concrete blocks; slabs, stairs, and walls are not affected.

Using version 'Railcraft_1.7.10-9.5.0.0"

WoodStone:


liach:
Will there be record on coreprotect plugin when breaking the block w/ crowbar? And the destruction can break blocks in those protected areas sometimes.

CovertJaguar:
@liach make a separate issue for missing block break events.

PR #542, opened 2015-03-11
Fix missing Block Sounds on SMP

PaperTriangle:
After some testing, this fixes the missing step-sounds on SMP Servers for blocks using the custom sound handler.

It also fixes the block-place sound for the same blocks that use this custom sound handler, however I have only been able to test this on locally hosted servers and am unsure if latency would stop the block-place sound from working.

onPlaySoundAtEntity was causing location issues when called from a client connected to an SMP Server. So it seemed redundant at this point and, considering sounds that it played seemed to get parsed through the onPlaySound handler anyways, I have removed it.


CovertJaguar:
Interesting...

PR #544, opened 2015-03-15
Fix for chunkload switch errors

ssotangkur:
You will need the pull request in the railcraft-api https://github.com/CovertJaguar/Railcraft-API/pull/7

I added the onBlockPlacedBy() handler to TileSwitchRouting since it wasn't detecting a pre-existing redstone signal when the switch is placed.

Switch tracks now implement getSwitchDevice() and it gets called every tick to find a valid the switch device. If one is found, its shouldSwitch() method is called (also on every tick).
Similarly, the switch device uses shouldSwitch() to determine what it's switch track is. It too checks the track for validity in its getter.


CovertJaguar:
Manually merged (line by line) to the correct branch. bb2c24a28d39cc2bf7e81faaa62072b10574c6fa
Closing.

PR #571, opened 2015-04-07
Made the Signal Block Relay Box more expensive

Vexatos:
Signal Block Relay Boxes are much cheaper than Block Signals but can be used to create Signal Blocks additionally to the other things they can do, making actual Block Signals almost redundant for anything but aesthetics. By making the Relay Box require a Block Signal, it gets more expensive and will probably only be used where really necessary.

If you don't like the change, just close this.


Vexatos:
Closing this in favour of a better change as discussed on IRC.

Issue #621, opened 2015-06-02
[Feature] Change direction of locomotive


DireTry20:
Add a new track that change the direction of a steam locomotive (sends it into reverse).

EDIT: There is currently a way to make it start going forwards but not backwards.

Vexatos:
Uuuh, control tracks set in the opposite direction will make the locomotive slow down until they make it turn, don't they?

Forecaster:
He means setting it to the reverse speed, not turning it around.

DireTry20:
thx, ill try harder in future

DireTry20:
I hope that this is not any trouble, but is there an ETA, im just curious about how dev happens

Forecaster:
Sorry, no ETA. Releases happen semi-sporadically, and since CJ has had a harddrive failure recently and lost a lot of stuff including some railcraft dev stuff it may be a while before the next one.

DireTry20:
it would doesn't need to be a new track but a change made to the limiter track

PR #642, opened 2015-06-27
Allow limiter track to run locomotives in reverse

cruck1000:
Per the idea of issue/suggestion #621 
Like the title says, this PR adds the ability for the limiter track to set locomotives in reverse.
It passed all my tests, then again, I didn't change much.
The texture is up for interpretation, I considered different colors, but in the end I think something like what I did would be good.
The code was **really** simple, changing 3 numbers and a texture :) (If I did it right)
(You write great code CJ)
Preview: image

Use it or don't, just trying to be helpful with my non-professional programming skills :)


Vexatos:
I'm not sure if this should be a thing...

cruck1000:
I wasn't quite sure either, I don't think there are an unlimited scale of uses, but for the people that actually want it, the feature could be really useful. It almost just seems like it should exist with the other functions of the limiter track.
Idk thou...

liach:
You might investigate that why dosen't CovertJaguar add this mode at first... Somewhat I feel something bad.

Forecaster:
I'm not really opposed. It's use it pretty much limited to allowing trains to automatically back into depos and the like, saving some tracks at the cost of  a few extra limiter tracks.

CovertJaguar:
I actually had textures for this at one point, but I lost them in the HDD crash. I just never got around to adding the functionality.

PR #669, opened 2015-07-28
adding foundation to village workshop

notabadminer:
Leaving foundation as cobblestone for now. This will be replaced with sandstone in desert villages.
The foundation could be brick if you want it to match the building construction, but I'll leave that up to you.


DZCreeper:
Perhaps concrete would make a better choice.

Logical in the fact that trains and the stations weight a lot and need strong foundations.

Illogical in the fact that we are talking about villagers making and forming concrete.

PR #692, opened 2015-09-22
Closes #598 Adds filter for named carts.

### LINKED ISSUE #598: [Enhancement] filters for loader/unloaders for named carts
KyeDuo:
> I would love it if you could filter by both cart type AND name if you used a named cart in the filter for the loaders/unloaders.
This would make for a very powerful but easy to use filter system.
A button in the GUI to either exclude or require the name on the cart/train would improve this as well.

http://www.railcraft.info/forum/?mingleforumaction=viewtopic&t=723.0#postid-2842

WoodStone:
The loaders can now filter named carts.
If you filter an unamed cart, then named carts will match.


CovertJaguar:
Closes #598

PR #693, opened 2015-09-22
More ore dictionary support

Adaptivity:
Chests and Crafting Tables were registered to support ganymedes01's WoodStuff mod. I've compiled new build with these changes successfully.
@CovertJaguar please, apply these changes by yourself because I can't sign CLA (getting "We're sorry, but something went wrong." error) and I don't really care who will apply these changes. These changes are simple and they are not code-based.


CovertJaguar:
Yeah, CLAHub is kind of sporadic. I'll consider this your statement of intent. ;)

On the plus side, there appears to be signs of life with CLAHub. It seems its gotten a new maintainer and partnered with another project. So maybe it will stop being so flaky.

PR #699, opened 2015-09-29
Update apis and introduced ic2classic api

### LINKED ISSUE #685: IC2 Classic Compatiblity request
I know CJ IC2 Classic is not a thing (at least public knowen and stuff) but could you implement some things support to it.

The only 1 thing which is really problematic with a lot of mods and that is there is never enough infromation transportet to the enet. I have manually injected some baseinformation and also a selflearning system created (after every start it does resets it data which i need to change) which kinda handles most of the stuff but not everything.

Could you implement this class also to your energySources?
https://github.com/TinyModularThings/IC2Classic/blob/master/ic2/api/energy/tile/IEnergySourceInfo.java
This class give the Enet all info which it needs to handle your Electric Source without problems.
How to use: Just return the value that it will emit max no matter what. (Please no Integer MaxValue. Real Values example: SteamTurbine produces max 200 EU. so you return 200 with it.)

I will add manual support to some of your generators but that solution is a dirty fix and i do not want to do everything manual for the end of my life...

Thanks for reading. Speiger

liach:
@Speiger is supposed to update this api on time.
Fix #685. Implemented that interface in energy receiver.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "api/thaumcraft/thaumcraft/api"]
 	path = api/thaumcraft/thaumcraft/api
 	url = https://github.com/Azanor/thaumcraft-api.git
+[submodule "ic2classic"]
+	path = api\\ic2classic
+	url = https://github.com/TinyModularThings/IC2Classic.git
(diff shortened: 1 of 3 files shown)


CovertJaguar:
I'm confused...why was the IC2 dependency removed?

liach:
Because the ic2classic api includes the ic2 exp api.

Speiger:
lol

liach:
Anyway I'm gonna fix it based on Speiger's instructions and try to get ic2 classic api from modmuss' maven instead of submoduling 8 hrs later.

liach:
I will remake this.

PR #701, opened 2015-09-30
Fix #300?

### LINKED ISSUE #300: Not to make minecart with command block show its gui when press lshift&right click with Crowbar
It shows its gui and it is not very easy to link carts. I hope you to change this. 
P.S. Thanks CovertJaguar! It's good to use Activator Track to disable the world anchor cart! But it didn't mentioned on the Wiki or anywhere else.

liach:
Add command block cart; Hope to fix crowbar issue, I might test it.
I guess crowbar prevents showing gui using forge MinecartInteractEvent. If yes then this must work.
Fix #300

### DIFF
--- a/src/main/java/mods/railcraft/common/util/misc/EntityIDs.java
+++ b/src/main/java/mods/railcraft/common/util/misc/EntityIDs.java
@@ -35,7 +35,7 @@ public class EntityIDs {
     public static final int CART_LOCO_STEAM_SOLID = 30;
     public static final int CART_LOCO_STEAM_MAGIC = 31;
     public static final int CART_LOCO_ELECTRIC = 32;
-    public static final int CART_COMMAND_BLOCK = 50;
+    public static final int CART_COMMAND_BLOCK = 40;
     public static final int CART_BORE = 50;
     public static final int CART_ENERGY_BATBOX = 87;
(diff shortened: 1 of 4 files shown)


src/main/java/mods/railcraft/common/util/misc/EntityIDs.java
```diff
@@ -35,6 +35,7 @@
     public static final int CART_LOCO_STEAM_SOLID = 30;
     public static final int CART_LOCO_STEAM_MAGIC = 31;
     public static final int CART_LOCO_ELECTRIC = 32;
+    public static final int CART_COMMAND_BLOCK = 50;
```
CovertJaguar: This is the same ID as the Tunnel Bore.

liach: Yes it should be 40; I am going to review.

AEnterprise: maybe it this could be converted to an enum to prevent conflicts like this in the future?

CovertJaguar: I'm not sure that IDs can be changed between versions without breaking save files. Otherwise I would be using an enum (I'm the enum king after all).

PR #702, opened 2015-10-04
Add command block cart into creative tab and fixed a cart summon bug

liach:
Sorry CJ. This time I've tested and it really works.

Other: Update gradle wrapper and fixed bug with idea textures (Believe it or not it is really fixed now)


CovertJaguar:
I'm hesitant to change the build script as it all works fine in IDEA for me.

src/main/java/mods/railcraft/common/modules/ModuleCore.java
```diff
@@ -213,6 +214,7 @@ public void initFirst() {
 
         // Items
         replaceVanillaCart(EnumCart.COMMAND_BLOCK, Items.command_block_minecart, "MinecartCommandBlock", 40);
+        Items.command_block_minecart.setCreativeTab(CreativeTabs.tabTransport);
```
CovertJaguar: This shouldn't be necessary. All minecart items default to the transport tab...

liach: But it does not show up in vanilla.

liach:
That line `idea.modules.inheritOutputDirs = true` works for me, and was suggested by Abrarsyed.

CovertJaguar:
well...I can try but the last time I tried something like this it broke release building.

CovertJaguar:
And when I redeployed my dev env back in August it worked out of the box without any special build targets in the IDE. So I'm not really clear on what problem is.

liach:
So should I test this with eclipse?

CovertJaguar:
Well you can if you want, but I don't really care about Eclipse. I'm using IntelliJ same as you.

liach:
Now does it hurt? Do I need to remove that line?

PR #703, opened 2015-10-04
Integrates with ic2classic better

liach:
Not shipping api but using `@Optional`. That might not hurt like last time.

### DIFF
--- a/src/main/java/mods/railcraft/common/plugins/ic2/TileIC2EmitterDelegate.java
+++ b/src/main/java/mods/railcraft/common/plugins/ic2/TileIC2EmitterDelegate.java
@@ -8,7 +8,9 @@
  */
 package mods.railcraft.common.plugins.ic2;
 
+import cpw.mods.fml.common.Optional;
 import ic2.api.energy.tile.IEnergySource;
+import ic2.api.energy.tile.IEnergySourceInfo;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraftforge.common.util.ForgeDirection;
 
@@ -16,7 +18,8 @@ import net.minecraftforge.common.util.ForgeDirection;
  *
  * @author CovertJaguar <http://www.railcraft.info/>
  */
-public class TileIC2EmitterDelegate extends TileIC2Delegate implements IEnergySource {
+@Optional.Interface(iface = "ic2.api.energy.tile.IEnergySourceInfo", modid = "IC2-Classic-Spmod")
+public class TileIC2EmitterDelegate extends TileIC2Delegate implements IEnergySource, IEnergySourceInfo {
 
     private final IEmitterDelegate delegate;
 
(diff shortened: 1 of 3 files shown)


src/main/java/mods/railcraft/common/plugins/ic2/TileIC2EmitterDelegate.java
```diff
@@ -16,7 +18,8 @@
  *
  * @author CovertJaguar <http://www.railcraft.info/>
  */
-public class TileIC2EmitterDelegate extends TileIC2Delegate implements IEnergySource {
+@Optional.Interface(iface = "ic2.api.energy.tile.IEnergySourceInfo", modid = "IC2-Classic-Spmod")
```
CovertJaguar: What about IEnergySource?

CovertJaguar:
I'm still not convinced that depending on the unofficial IC2 classic API is a great idea....

CovertJaguar:
Ok...so I looked at this and the possibility of removing the IC2 files from the jar....and supporting IC2 classic pretty much makes that impossible. The @Optional framework doesn't support interfaces/methods that can be present in two mods....

bookerthegeek:
Can a mod have two internal names? That way maybe ic2 classic could add an additional name you could call against, why still allowing it to show as ic2?

CovertJaguar:
Well the problem is actually that all the base IC2 stuff would disappear if IC2 wasn't installed, even if IC2-classic is installed. IC2-classic would need to report as IC2 as well.

CovertJaguar:
I'm told the function you want implemented is already provided by EnergyNet.instance.getPowerFromTier(getSourceTier()). So unless you give me a really good reason for implementing this redundant interface, I'm probably going to just remove the IC2-classic API integration.

liach:
@Speiger had this idea first. I think he might be able to explain or you can remove this change.

Speiger:
@CovertJaguar Yeah the Function does what you say. But the problem here simply is it is not detailed enough for classic.
Classic builds paths for the Enet instead of adding a node (Exp). So its calculating all valid paths based on the power it is outputing. Now lets say your machine outputs 6 EU but has a tier of 32EU. Then it generates simply more paths that it could not handle, because of to much powerloss, it would aventually remove the paths which are not there but they system would always generate more paths as it should. Basicly this function i asked you to implement is simply a helper for performance. Because if you have big systems that could cause huge lag (like if you trave 200 blocks for each receiver and you have 6 differend receivers). And the problem with my Selflearning system is that it can only learn up... To get the upper limits...

Is that explaining enough?

CovertJaguar:
The function is a simple lookup, I fail to see why this must be implemented on my end. The function has no input and always maps to same values...it is not needed.

liach:
Btw what's the difference between `double getOfferedEnergy()` and `int getMaxEnergyAmount()` ? I think you speiger can totally use the previous one.

Speiger:
@liach i can use that one right but that number will change based on the amount which is stored.
And do you know a generator which is full powered when it is placed? No.
And the Enet is build only paths if new EnergyTiles are placed. And if the generator has no charge the max distance is then 0 because it has 0 eu to offer.

CovertJaguar:
double getOfferedEnergy() is how much power the block can provide to the system at any given moment.

PR #731, opened 2015-11-17
Make EntityCartFiltered use the DataWatcher for the filter ItemStack

maruohon:
Note that the client-side PhantomInventory still won't be in sync until
a player opens the Container and then the filter ItemStack will get
synced via the Container.
But the PhantomInventory shouldn't currently be used for anything
other than backing the Container on the client side anyway...

Rendering and other "external" stuff uses the
EntityCartFiltered#getFilterItem() method, which now returns the filter
ItemStack directly from the DataWatcher.


CovertJaguar:
Looks good!

PR #738, opened 2015-11-26
Added new condition for checking locomotive type.

Cream-Tea:
Added new routing table condition to check what the type of locomotive is.
Returns true if the locomotive type matches one of the simple to use strings "electric", "steam", or "none". (Can accept "steam_magic", but wasn't sure if it's needed)
Using "none" only returns true if there is NO locomotive.
But for the others it's fairly simple "electric" only returns true if the locomotive is electric.


CovertJaguar:
Looks good.

PR #744, opened 2015-11-29
New track for locomotives.

Cream-Tea:
If you haven't yet, here's the concept video:
https://www.youtube.com/watch?v=Zaxqc7XZ3P8

That shows you how it would work, the track is that green looking limiter track.
I also show you that it has a skin for both powered and unpowered as it must be powered.

It has worked fully when I tested it, as you see that the testing in the video.


liach:
Rebase and squash.

Cream-Tea:
I've never used the data watcher from forge before, If you think it will benefit then sure. From what I know about it, it just reduces the amount of data sent to the client if it doesn't need certain information.

liach:
Data watcher is from vanilla.
I assume the data watcher could add link information to all the minecarts, so we do not need separate fields that only works with the locomotive.

Cream-Tea:
I intended the track to only work with locomotives, since locomotives have the reverse capability etc...
So personally I don't think it's a problem, unless we want it to work with normal minecarts. But then we'd have to check whether it's a minecart because right now it automatically sets the train to the slowest forward speed when it couples up after the track has tagged it as "ready".

CovertJaguar:
No, what is should be is a NBT data entry in the EntityData. See how "HighSpeed" works in MinecartHooks. 

https://github.com/CovertJaguar/Railcraft/blob/master/src/main/java/mods/railcraft/common/carts/MinecartHooks.java#L291

liach:
That entity data is a functional hook [added by forge](https://github.com/MinecraftForge/MinecraftForge/blob/master/patches/minecraft/net/minecraft/entity/Entity.java.patch#L139). Does any minecart belong to a train?

CovertJaguar:
@liach yes all minecarts belong to a Train.

liach:
Then I suggest to add a boolean in the train instance to mark whether it should auto connect next time.

CovertJaguar:
Thoughts on how this would work as a Coupler Track mode rather than a new Track?

Cream-Tea:
Well if we made a state for the coupler track, which can be changed to using the crowbar like some of the other tracks do. It could be like an "Auto-Coupling" track.

Thing is the main function of this "auto-coupling" track would be that it must keep the idea of "locomotive backs into a cart" --> "cart links with loco automatically" --> "loco starts going forward at shunting speed".

The last bit is essential because if it just automatically couples up but doesn't move, it'll have no point since it will just sit there and somebody will have to start it moving.

So as long as you have those three steps then all would be good, would you like me to start working on modifying the "Coupler Track" to add this functionality?

src/main/java/mods/railcraft/common/blocks/tracks/EnumTrack.java
```diff
@@ -82,7 +82,8 @@
     ELECTRIC_JUNCTION(Module.TRACKS_ELECTRIC, 1, 0, "electric.junction", 8, TrackElectricJunction.class),
     ELECTRIC_SWITCH(Module.TRACKS_ELECTRIC, 4, 0, "electric.switch", 8, TrackElectricSwitch.class),
     ELECTRIC_WYE(Module.TRACKS_ELECTRIC, 2, 0, "electric.wye", 8, TrackElectricWye.class),
-    FORCE(Module.ELECTRICITY, 1, 0, "force", 1, TrackForce.class),;
+    FORCE(Module.ELECTRICITY, 1, 0, "force", 1, TrackForce.class),
+    AUTOLOCOCOUPLER(Module.LOCOMOTIVES, 2, 1, "autolococoupler", 8, TrackAutoLocoCoupler.class);
```
liach: make it into 
```java
    AUTOLOCOCOUPLER(Module.LOCOMOTIVES, 2, 1, "autolococoupler", 8, TrackAutoLocoCoupler.class),
    ;
```
for less patch in the future.

CovertJaguar:
Sure, look at NextGenLoading for the most applicable example of track modes.

PR #747, opened 2015-12-05
Implementation for issue #561

### LINKED ISSUE #561: [Enhancement] Allow Switch Motors to ignore redstone signals
Currently a normal Switch Motor will respond to both a signal controller and redstone signals around it. It obviously will not respond to aspects if nothing has been connected to it, but there is no way to have it ignore interference from surrounding redstone signal sources even if you'd only want it to react to it's assigned signal controller.

I'd like a button in the gui that allows you to toggle it's ability to react to redstone on and off.

The only solution currently is to spread things out in an ugly manner to remove the sources of the interfering signals.

wshadow:
May not be the best implementation (a lot of code copied), but:
* Allows the switch motor to ignore redstone signals.
* Also adds a button to the gui to toggle that behavior.
* Existing switch motors (should) default to react to a redstone signal.

closes #561 

also see https://github.com/CovertJaguar/Railcraft-Localization/pull/303 for the German and English localization for the new button


CovertJaguar:
This one will need in-game review I think. I'll get to it before next release.

PR #784, opened 2016-03-11
Classic Compatiblity Patch...

Speiger:
-Added: MFSU Cart (only for IC2 Classic)
-Added: MFE Cart Adjust to IC2 Classic when its loaded
-Changed: CESU Cart get not loaded when IC2 Classic is installed.
-Changed: CESU Cart shows as MFSU If it get loaded for Any Reason if Classic installed
-Missing: MFSU Cart (When i make Textures people die so yeah)

If anything is wrong. Suggest something

### DIFF
--- a/src/main/java/mods/railcraft/common/carts/EntityCartEnergyMFE.java
+++ b/src/main/java/mods/railcraft/common/carts/EntityCartEnergyMFE.java
@@ -10,6 +10,7 @@ package mods.railcraft.common.carts;
 
 import net.minecraft.item.ItemStack;
 import net.minecraft.world.World;
+import mods.railcraft.common.modules.ModuleIC2;
 import mods.railcraft.common.plugins.forge.LocalizationPlugin;
 import mods.railcraft.common.plugins.ic2.IC2Plugin;
 
@@ -19,8 +20,11 @@ import mods.railcraft.common.plugins.ic2.IC2Plugin;
  */
 public class EntityCartEnergyMFE extends EntityCartEnergy {
 
+	boolean classic;
+	
     public EntityCartEnergyMFE(World world) {
         super(world);
+        classic = ModuleIC2.classic;
     }
 
     public EntityCartEnergyMFE(World world, double d, double d1, double d2) {
@@ -41,17 +45,17 @@ public class EntityCartEnergyMFE extends EntityCartEnergy {
 
     @Override
     public int getTier() {
-        return 3;
+        return classic ? 2 : 3;
     }
 
     @Override
     public int getCapacity() {
-        return 4000000;
+        return classic ? 600000 : 4000000;
     }
 
     @Override
     public int getTransferLimit() {
-        return 512;
+        return classic ? 128 : 512;
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


src/main/java/mods/railcraft/common/modules/ModuleIC2.java
```diff
@@ -46,7 +49,7 @@ public void printLoadError() {
 
     @Override
     public void initFirst() {
-
+    	classic = Loader.isModLoaded("IC2-Classic-Spmod");
```
CovertJaguar: Put this in IC2Plugin and get rid of all the local booleans?

CovertJaguar:
@Speiger You need to sign the CLA

src/main/java/mods/railcraft/common/carts/EntityCartEnergyMFE.java
```diff
@@ -19,8 +20,11 @@
  */
 public class EntityCartEnergyMFE extends EntityCartEnergy {
 
+	boolean classic;
```
liach: I suggest to refer to the static boolean instead of making instance fields.

src/main/java/mods/railcraft/common/carts/EntityCartEnergyMFSU.java
```diff
@@ -0,0 +1,55 @@
+package mods.railcraft.common.carts;
+
+import ic2.api.item.IC2Items;
+import mods.railcraft.common.plugins.forge.LocalizationPlugin;
+import net.minecraft.item.ItemStack;
+import net.minecraft.world.World;
+
+public class EntityCartEnergyMFSU extends EntityCartEnergy {
```
liach: What if it switched back to a non ic2 classic environment? Show this as cesu?

Speiger: It should despawn the second the world get loaded... If Classic is installed the CESU Cart does not get registered in the EntityList. And if it is not installed the MFSU Cart does not get registered... You will get a lot of (If used) Skipping Entities but that should be it...

src/main/java/mods/railcraft/common/modules/ModuleIC2.java
Speiger: ok

Speiger:
@CovertJaguar while i did the changes i discovered something interesting... Could i add the IMachineUpgradeItem / IMachine to your mod simply to support the new Upgrades i have?
(EnergyMultiplier Upgrade and also add support for Classic that the Lapotron Upgrade works)?

Speiger:
@CovertJaguar updated.

CovertJaguar:
Regarding IMachine, I'd entertain a separate PR sure.

src/main/java/mods/railcraft/common/modules/ModuleIC2.java
CovertJaguar: Generally, maintainability is 10x more important than speed.

Speiger:
@CovertJaguar that was my idea... But i just wanted to hear if you agree or if i should leave it away...

Speiger:
@CovertJaguar + @liach another patch. Simply some recipe issues if classic is installed...

liach:
@Speiger I suggest you to squash commits and fix your email for your commit in order to show these change as your contribution.

PR #785, opened 2016-03-12
Merge New Track Idea Into Coupler Track.

Cream-Tea:
As the title suggests I've used the new track idea but implemented is a state in the coupler track.

Video demonstration:
https://www.youtube.com/watch?v=o70Lwg-LugQ


.metadata/.plugins/org.eclipse.core.resources/.root/.indexes/history.version
```diff
@@ -0,0 +1 @@
+
```
Vexatos: Something something `.gitignore`

PR #796, opened 2016-03-29
Fix a lot of compile errors in 1.8.9 port

AlexIIL:
[ main | [api](https://github.com/CovertJaguar/Railcraft-API/pull/10) ]

I hope this helps :)

I haven't finished yet, and my current plan is to get all server-side stuff working and functional before making everything render properly. I might be moving in the wrong direction for some things (I am just doing a straight port rather than anything special as I don't know the codebase well.)

Before this is merged (or just general TODO stuffs)
- [ ] Fix all compile errors
- [ ] Make sure all 1.7.10 blocks + items migrate (unlike BC I will be careful at not breaking metadata)
- [ ] Fix renderers for non-track blocks

This wants your opinion on how I implement it:
- [ ] Properly customizable track models
      - Essentially how you can change them, this really wants to be investigated fully later.


src/main/java/mods/railcraft/client/particles/ParticleHelper.java
```diff
@@ -35,42 +36,48 @@
...
+//        if (sideHit == SOUTH)
+//            pz = z + block.getBlockBoundsMaxZ() + b;
+//
+//        if (sideHit == WEST)
+//            px = x + block.getBlockBoundsMinX() - b;
+//
+//        if (sideHit == EAST)
+//            px = x + block.getBlockBoundsMaxX() + b;
+        // TODO 1.8.9 port make sure this si the correct function and it works!
+        effectRenderer.addBlockHitEffects(pos, sideHit);
```
CovertJaguar: I doubt it, it always used the wrong texture before (hence the override).

src/main/java/mods/railcraft/client/particles/ParticleHelper.java
```diff
@@ -89,23 +96,25 @@ public static boolean addHitEffects(World world, Block block, MovingObjectPositi
...
-                    if (callback != null)
-                        callback.addDestroyEffects(fx, world, pos, state);
-                    effectRenderer.addEffect(fx.func_174845_l());
-                }
-            }
-        }
+//        if (block != WorldPlugin.getBlock(world, pos)) return true;
+        // TODO 1.8.9 port make sure this si the correct function and it works!
+        effectRenderer.addBlockDestroyEffects(pos, state);
+//        byte its = 4;
```
CovertJaguar: Again, probably wrong texture.

AlexIIL: Because of the way models work now it *must* be the correct texture as that's the only texture we have access to

src/main/java/mods/railcraft/common/blocks/machine/TileMachineItem.java
```diff
@@ -16,7 +16,9 @@
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 
-public abstract class TileMachineItem extends TileMachineBase implements IInventory {
+public abstract class TileMachineItem extends TileMachineBase 
+// FIXME REMOVE and replace with capabilities
+implements IInventory {
```
CovertJaguar: I'd rather not. I don't see any benefit.

src/main/java/mods/railcraft/client/particles/ParticleHelper.java
CovertJaguar: Then it would be better to remove the override at the block level. But I still doubt that it will use the correct texture and yes there are ways to access textures directly from what I saw in my brief inspection of this code.

src/main/java/mods/railcraft/common/blocks/machine/TileMachineItem.java
AlexIIL: IInventory is considered obsolete by forge now. However it would probably be easy to provide an inventory capability wrapper that would satisfy everyone.

CovertJaguar:
@AlexIIL this is too much...I can't process this. Especially, forcing things to compile without fixing underlying issues will just make things harder later. D=

CovertJaguar:
I wish I could do a line by line or file by file merge. =/

AlexIIL:
I made this PR too early in my port process, I don't really want to merge this in now as I want to test everything before hand.

CovertJaguar:
Honestly I'd rather you split it up and made MUCH smaller PRs.

As I said, I can't process this and it removes my ability to tell which code is still broken, while leave it still broken.

src/main/java/mods/railcraft/common/blocks/machine/TileMachineItem.java
CovertJaguar: Then forge is short sighted, IItemHandler is too restrictive for many use cases.

AlexIIL:
Fair enough. I could attempt to split up my work, or I could continue until everything works and give you a ported version of railcraft. (I chose the latter option with BuildCraft, and the former only works if you are available all the time to check everything)

CovertJaguar:
I am actively working on the port myself. I've been methodically rewriting the broken portions of the code. I don't really care if it compiles until that stage is done. A lot of what you are doing is just making it compile without actually fixing underlying issues. This removes my ability to tell which code needs rewriting still. I just don't know how merge something this massive an encompassing into my workflow.

AlexIIL:
Yeah, I'm forgetting that I'm not the only one porting it. Also, the only underlying issues I am leaving in are rendering, which is a whole different beast. To make it work it really needs to have the server-side functionality working fully before attempting to change io.

src/main/java/mods/railcraft/common/core/Railcraft.java
```diff
@@ -210,7 +213,7 @@ public void serverCleanUp(FMLServerStoppingEvent event) {
     public void missingMapping(FMLMissingMappingsEvent event) {
         for (FMLMissingMappingsEvent.MissingMapping mapping : event.get()) {
             if (mapping.type == GameRegistry.Type.BLOCK) {
-                Block block = GameRegistry.findBlock(MOD_ID, MiscTools.cleanTag(mapping.name));
+                Block block = GameRegistry.findBlock(MOD_ID, mapping.resourceLocation.getResourcePath());
```
CovertJaguar: This is completely wrong. I don't know what the correct solution is yet, but I know this wrong. The entire point of this remapping was to clean the garbage out of the tag. It might be necessary to remove it completely.

src/main/java/mods/railcraft/common/core/RailcraftConfig.java
```diff
@@ -564,12 +564,12 @@ private static void loadBlocks() {
     }
 
     private static void loadBlockProperty(String tag) {
+        System.out.println("loadBlockProperty("+tag+")");
         Property prop = configBlock.get(CAT_BLOCKS, tag, true);
         enabledBlocks.put(tag, prop.getBoolean(true));
     }
 
     private static void loadBlockFeature(String tag) {
-        tag = MiscTools.cleanTag(tag);
```
CovertJaguar: These are necessary.

src/main/java/mods/railcraft/common/items/ItemCircuit.java
```diff
@@ -31,10 +31,10 @@ public ItemCircuit() {
 
     @Override
     public void initItem() {
-        for (EnumCircuit circuit : EnumCircuit.VALUES) {
-            ItemStack stack = new ItemStack(this, 1, circuit.ordinal());
-            RailcraftRegistry.register(stack);
-        }
+//        for (EnumCircuit circuit : EnumCircuit.VALUES) {
```
CovertJaguar: Why?

src/main/java/mods/railcraft/common/plugins/forge/LootPlugin.java
```diff
@@ -70,7 +70,7 @@ public static void addLoot(RailcraftItem item, IItemMetaEnum meta, int minStack,
 
     public static void addLootUnique(RailcraftItem item, IItemMetaEnum meta, int minStack, int maxStack, Type type) {
         ItemStack stack = item.getStack(meta);
-        addLoot(stack, minStack, maxStack, stack.getUnlocalizedName(), type.locations);
+        addLoot(stack, minStack, maxStack, item.getBaseTag(), type.locations);
```
CovertJaguar: I wrote this function the other day specifically because item.getBaseTag() is too generic.

src/main/java/mods/railcraft/common/util/misc/MiscTools.java
```diff
@@ -50,10 +50,6 @@ public static void registerTrack(EnumTrack rail) {
             }
     }
 
-    public static String cleanTag(String tag) {
-        return tag.replaceAll("[Rr]ailcraft\\p{Punct}", "").replaceFirst("^tile\\.", "").replaceFirst("^item\\.", "");
```
CovertJaguar: Ok...now I'm confused. I'm pretty sure that removing this will break pretty much everything.

src/main/java/mods/railcraft/common/core/Railcraft.java
hea3ven: Just in case, to save you debugging headaches, this event is currently not working: https://github.com/MinecraftForge/MinecraftForge/issues/2491

AlexIIL: This bug is currently stopping BuildCraft from migrating 1.7.10 worlds and migrating fluids after I refactored the way that fluids were registered.

src/main/java/mods/railcraft/common/util/misc/MiscTools.java
AlexIIL: All of my registry changes were pretty much done as the game would immediately crash when you attempted to register an item twice. I think this is stuff that I don't understand properly so I won't change any block or item registration stuff in future and leave it to you to sort out.

AlexIIL: Oh, and there is a separate "getRegistryName" and "setRegistryName" so this is not necessary for registry stuff. (So we are discouraged from using getUnlocalisedName for registry stuff.)

CovertJaguar: Yeah the registry name stuff is on my todo list, but the tag cleaning is used everywhere from localization to config files.

AlexIIL:
Closing this as I'll put everything that I fix properly in separate small PR's

PR #799, opened 2016-03-30
[1.8.9] Fix sounds, update RF API to latest.

AlexIIL:
Also fix a small mistake in my earlier PR.

This PR is compatible with my other PR's (You can merge them in any order, or reject any of them and all the others will still function)

### DIFF
--- a/src/main/java/mods/railcraft/client/sounds/RCSoundHandler.java
+++ b/src/main/java/mods/railcraft/client/sounds/RCSoundHandler.java
@@ -13,13 +13,12 @@ import mods.railcraft.common.util.sounds.IBlockSoundProvider;
 import mods.railcraft.common.util.sounds.SoundRegistry;
 import net.minecraft.block.Block;
 import net.minecraft.block.Block.SoundType;
+import net.minecraft.block.state.IBlockState;
 import net.minecraft.client.audio.PositionedSoundRecord;
 import net.minecraft.util.BlockPos;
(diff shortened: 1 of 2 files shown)


src/main/java/mods/railcraft/client/sounds/RCSoundHandler.java
```diff
@@ -59,18 +56,19 @@ public void onPlaySound(PlaySoundEvent event) {
...
-        Block block = world.getBlockState(new BlockPos(x, y, z)).getBlock();
+    private SoundType getBlockSound(World world, BlockPos pos) {
+        IBlockState state = world.getBlockState(pos);
+        Block block = state.getBlock();
         if (block instanceof IBlockSoundProvider)
-            return ((IBlockSoundProvider) block).getSound(world, x, y, z);
+            return ((IBlockSoundProvider) block).getSound(world, pos);
         else {
-            int meta = world.getBlockMetadata(x, y, z);
+            int meta = block.getMetaFromState(state);
```
CovertJaguar: This needs to change but I can take care of that after merging.

AlexIIL: All block meta comparisons -> BlockState comparisons? Or just change the sound handler?

CovertJaguar: I probably need to kill the sound registry and only use the interface.

Issue #857, opened 2016-07-22
Vanilla Cart Replacement Broken


CovertJaguar:
It appears that two carts are once against being spawned for every item.

PR #861, opened 2016-08-08
Fix ore crash and #857

liach:
Also fixes some issue with build script and compiling errors (bad java),
using Java 1.8_102.

Both fix are just one line of code. I've tested and they are effective. The crash for ore:
https://gist.github.com/liach/39b6100a4960e3d3c2dcbd4806e1dd76

Ref #857 Cart Replacement Broken


CovertJaguar:
Sorry, you put too many things into one PR and my local changes conflicted (track stuff). You'll need to rebase.

liach:
@CovertJaguar Conflict fixed.

PR #872, opened 2016-08-20
Crowbar events

### LINKED ISSUE #547: Throw some forge events when crowbar breaking blocks
It seems that the breaking function of the crowbar doesn't call any event. Maybe call some would improve compatibility.

liach:
Fixes #547


liach:
Is this okay?

CovertJaguar:
I have no plans to release another 1.7 version at this time.

liach:
I agree that we need to focus on minecraft 1.10.2. I'd just leave this pr here as it is, until a stable version of Railcraft 10 is published probably.

liach:
May I build a jar from this patch and use in a server and its mod pack? I will add a password lock to its mod pack. I want your permission.

AEnterprise:
i think you could get away with just putting this on the server, as far as i see this is code that is/should only be executed on the server side

PR #880, opened 2016-08-27
Fixed cart linking issue

### LINKED ISSUE #868: [1.10.2] Cannot link cart
When train module is enabled you cannot link carts for some reason.

liach:
Fixes #868. Unluckily you need to close that issue by hand because I forgot to reference to it in my commit.
Test results:
(I made a link previously when I had a lot of code for debug infos, so at first I broke a link)

Well unfortunately the distance between carts seems to be incorrect. We shall investigate later.


liach:
So silly am I that I just figured out I can actually close that issue once you merge it, since I opened it.

liach:
wait do not merge yet

liach:
Okay now.

Issue #896, opened 2016-09-17
Errorous village generation


liach:
![2016-09-17_16 41 12](https://cloud.githubusercontent.com/assets/7806504/18611668/a2332776-7cf5-11e6-8b7d-6ff16bd898ac.png)
Some blocks which are supposed to be stone bricks became air and torches falls. The machine can be disregarded for now though.

liach:
Fix found, will send a trivial pr soon

PR #897, opened 2016-09-18
Fixed roof generation

liach:
Used a safer way to get cyan glass block state.
The MCP mapping for a parameter is wrong, so that the roof does not generate.
Fixed torch generations as well.
The loot table is not fixed, though.
Screenshot:


liach:
Btw ref #896

CovertJaguar:
Oh good. I wasn't sure how to fix this one.

PR #905, opened 2016-09-28
Fixed #901

### LINKED ISSUE #901: [1.7.10] Dupe with Cart Dispenser
https://www.youtube.com/watch?v=CYoFOP5OLCI

liach:
When forge replace the normal item spawned with a custom entity item, `spawnEntityInWorld` returns false and the custom item is dropped without consuming the item in the cart dispenser.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/machine/gamma/TileDispenserCart.java
+++ b/src/main/java/mods/railcraft/common/blocks/machine/gamma/TileDispenserCart.java
@@ -8,18 +8,6 @@
  */
 package mods.railcraft.common.blocks.machine.gamma;
 
-import java.io.DataInputStream;
-import java.io.DataOutputStream;
-import java.io.IOException;
-import net.minecraft.entity.item.EntityItem;
-import net.minecraft.entity.item.EntityMinecart;
-import net.minecraft.entity.player.EntityPlayer;
-import net.minecraft.inventory.IInventory;
-import net.minecraft.item.ItemMinecart;
-import net.minecraft.item.ItemStack;
-import net.minecraft.nbt.NBTTagCompound;
-import net.minecraft.util.IIcon;
-import net.minecraftforge.common.util.ForgeDirection;
 import mods.railcraft.api.carts.CartTools;
 import mods.railcraft.api.core.items.IMinecartItem;
 import mods.railcraft.common.blocks.machine.IEnumMachine;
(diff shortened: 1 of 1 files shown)


liach:
Note: If you want a hotfix release for Minecraft 1.7, I can provide more fixes on mow cart griefing issues.

src/main/java/mods/railcraft/common/blocks/machine/gamma/TileDispenserCart.java
```diff
@@ -125,7 +127,7 @@ public void onPulse() {
                             float rx = MiscTools.getRand().nextFloat() * 0.8F + 0.1F;
                             float ry = MiscTools.getRand().nextFloat() * 0.8F + 0.1F;
                             float rz = MiscTools.getRand().nextFloat() * 0.8F + 0.1F;
-                            EntityItem item = new EntityItem(worldObj, x + rx, y + ry, z + rz, cartStack);
+                            Entity item = cartStack.getItem().createEntity(worldObj, new EntityItem(worldObj, x, y, z, cartStack), cartStack);
```
CovertJaguar, line 130: What does this function do? SpawnEntityInWorld?

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/blocks/machine/gamma/TileDispenserCart.java
liach: The problem is actually [here](https://github.com/liachmodded/Railcraft/blob/4991b96cfe022c4a99642d5716f3d83b5c54ab78/src/main/java/mods/railcraft/common/blocks/machine/gamma/TileDispenserCart.java#L135). When forge replaces the vanilla `EntityItem` with mod custom item entity, it returns false, causing the slot of item stack in the cart dispenser not clearing the slot. So I jsut let Railcraft spawn the custom mod item entity at first, so that the call there will not return a confusing result.

CovertJaguar: You haven't tested this have you? The default implementation of Item.createEntity() just returns null. The normal outcome will be a NPE.

CovertJaguar: The correct solution appears to be to just remove the if statement.

liach: I just tested it with firestone.

liach:
> The correct solution appears to be to just remove the if statement.

Then I will close this.

PR #945, opened 2016-10-16
Fix Routing Table, Golden Ticket editing

### LINKED ISSUE #944: Routing Table, Golden Ticket edits don't propagate to server
Editing one of these will appear to take effect on the client, but the change will be lost when the item is dropped on the ground, put in a Routing Detector, etc.

mconwa01:
Fixes #944.

(I'll say up front that I'm new at this, so if there's anything that looks wrong, it probably is.)

It looks like `player.getActiveItemStack()` (PacketHandler.java) is returning `null` on the server, so `PacketCurrentItemNBT.readData()` returns early. I think the attribute behind `getActiveItemStack()` is getting reset on the next tick after the right click that triggers the edit, so it's `null` by the time the edit GUI is closed and the packet reaches the server. (Vanilla books appear to go after the main hand slot only, so they fail when editing a book in the off hand... [MC-84005](https://bugs.mojang.com/browse/MC-84005))

I check the player's hands in the case that `PacketCurrentItemNBT.currentItem` is null. This feels a little bit hacky, but if my theory is correct, I don't see a place on the server that would remember which item the player "used" last.

### DIFF
--- a/src/main/java/mods/railcraft/common/util/network/PacketCurrentItemNBT.java
+++ b/src/main/java/mods/railcraft/common/util/network/PacketCurrentItemNBT.java
@@ -18,8 +18,8 @@ import org.apache.logging.log4j.Level;
 
 import javax.annotation.Nullable;
 import java.io.IOException;
+import java.util.stream.Stream;
 
-//TODO: dual welding breaks this
 public class PacketCurrentItemNBT extends RailcraftPacket {
 
     private final EntityPlayer player;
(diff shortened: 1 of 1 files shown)


liach:
I do not think this is the correct fix. Does this look like an issue with threads?

mconwa01:
You could be right--I don't yet see evidence that it's a threading issue, but I don't understand the dual-wielding rat's nest enough yet either. What I see is that
- The main branch has the issue, both pre- and post- 1fe66db.
- The issue seems to be caused by the fact that `PacketCurrentItemNBT.currentItem` is `null` on the server in `readData`, at least pre- 1fe66db (haven't directly checked post).
- My change fixes the issue, at least in normal usage, both pre- and post- 1fe66db.
  - There may well be edge cases, e.g. I suspect that if the player is editing a routing table in the off hand and another routing table lands in the main hand in the meantime, the edit may incorrectly go to the main hand item.
- Vanilla writeable books aren't much help as an example, as they just go for the main hand (see [MC-84005](https://bugs.mojang.com/browse/MC-84005)).

src/main/java/mods/railcraft/common/util/network/PacketCurrentItemNBT.java
```diff
@@ -39,15 +39,22 @@ public void writeData(RailcraftOutputStream data) throws IOException {
     @Override
     public void readData(RailcraftInputStream data) throws IOException {
         try {
-            ItemStack stack = data.readItemStack();
+            ItemStack stack = data.readItemStack(), currentItem;
```
CovertJaguar, line 42: Why are we shadowing currentItem?

CovertJaguar (commented):
Has this been tested?

src/main/java/mods/railcraft/common/util/network/PacketCurrentItemNBT.java
mconwa01: Just because it seemed like the correct fix would be to get the right `currentItem` from the start (in the constructor), but I couldn't find a way to do it without the one in the incoming packet, so I opted to keep the diff small for now. If this is the solution that is used, I agree it would be best to pick a new name (and/or make wider changes, since the `getActiveItemStack()` call on the server isn't helping).

CovertJaguar: Yeah, the top level call seemed kind of redundant and awkward to me too but I'm not really sure how to do it differently.

Issue #940, opened 2016-10-14
Add analog output to Routing Detector


mconwa01:
The important bit is to allow a single Routing Detector to distinguish between more than two states. It would pair well with the Analog Signal Controller Box for moderately complex builds and clean up places where we'd currently need multiple Detectors around the same Track.

This change should be possible without upsetting existing builds by modifying the routing table language to operate on nonnegative integers instead of booleans, but otherwise remain the same.
- Existing TRUE values would be replaced by REDSTONE_MAX (or somesuch--whatever would preserve existing behavior, even with mods that allow redstone signals > 15), FALSE would be 0.
- Leaf conditions would output one of these two values.
- AND and OR would treat 0 as FALSE and anything else as TRUE, but would evaluate to the same value as their last-evaluated argument (assuming short-circuit evaluation).
  - (AND 0 1) = 0
  - (OR 0 1) = 1
  - (AND TRUE 5) = 5
  - (OR TRUE 5) = TRUE
- NOT would return TRUE for 0, and 0 for anything else.
- Constant numeric values would be added as leaf nodes. It might be convenient to add TRUE (REDSTONE_MAX) and FALSE constants as well.

Example:

```
AND
Dest=foo
15
AND
Dest=bar
14
```

This would output a redstone signal of 15 for a destination starting with `foo`, 14 for `bar`, and 0 for anything else (no train present or other destination).

In the future, it might also be convenient to add condition nodes that output analog signals, or edit existing ones (e.g. instead of or in addition to Redstone=<true/false>, let Redstone read the current redstone input).

This need only affect Routing Detectors; Routing Switch Motors could just treat 0 as false and anything else as true. Another option would be to add a dedicated Analog Routing Detector, but according to #856, this would be less desirable.

[...]

PR #947, opened 2016-10-16
Add analog behavior to the Routing Detector

mconwa01:
Implements #940.
- I use -1 as a sentinel for "maximum value" to keep `RoutingLogic` independent of the range of redstone power levels. This happens to bleed through to the user (entering -1 will have the same effect as TRUE, while any other negative number will raise a `RoutingLogicException` for an "Invalid Constant"). Do you mind?
- This still needs explanation in the routing table manual and a language file entry for the "Invalid Constant" message.


src/main/java/mods/railcraft/common/blocks/detector/types/DetectorRouting.java
```diff
@@ -105,8 +105,9 @@ public int testCarts(List<EntityMinecart> carts) {
             if (routingController.getButtonState() == RoutingButtonState.PRIVATE)
                 if (!getOwner().equals(CartToolsAPI.getCartOwner(cart)))
                     continue;
-            if (logic.matches(this, cart))
-                return FULL_POWER;
+            int value = logic.evaluate(this, cart);
+            if (value != 0)
```
liach: Unnecessary

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
```diff
@@ -204,8 +222,8 @@ public NOT(Condition a) {
         }
 
         @Override
-        public boolean matches(IRoutingTile tile, EntityMinecart cart) {
-            return !a.matches(tile, cart);
+        public int evaluate(IRoutingTile tile, EntityMinecart cart) {
+            return valueForBool(a.evaluate(tile, cart) == 0);
```
liach:
``` java
return 16 - a.evaluate(tile, cart);
```

mconwa01: That's a different behavior, though worth considering. We'd have to nix the special -1 though. My original thought was that most devices treat any power level as "on", so the inverse should be "off" (i.e. `== 0` behaves like a redstone torch).

src/main/java/mods/railcraft/common/blocks/detector/types/DetectorRouting.java
mconwa01: This preserves the original behavior of moving to the next cart in `carts` if the table does not match (plus we don't have to reevaluate on the following line). Though I see now that the result depends on the order of `carts` if multiple carts trigger different nonzero signal levels. We could take the maximum value (or something like that) in order to make it consistent.

liach: Ok it is for each cart.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
```diff
@@ -93,17 +93,18 @@ private EntityMinecart getRoutableCart(EntityMinecart cart) {
...
         EntityMinecart controllingCart = getRoutableCart(cart);
         if (controllingCart == null)
-            return false;
+            return 0;
         for (Condition condition : conditions) {
-            if (condition.matches(tile, controllingCart))
-                return true;
+            int value = condition.evaluate(tile, controllingCart);
+            if (value != 0)
+                return value;
```
liach: Should it be first match or max here?

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
```diff
@@ -164,8 +176,14 @@ public ToolTip getToolTip() {
 
     private abstract class Condition {
```
liach: Should this be turned into an interface for convenience?

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
liach: I think you should replace -1 with `PowerPlugin.FULL_POWER`

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
```diff
@@ -220,8 +238,8 @@ public AND(Condition a, Condition b) {
         }
 
         @Override
-        public boolean matches(IRoutingTile tile, EntityMinecart cart) {
-            return a.matches(tile, cart) && b.matches(tile, cart);
+        public int evaluate(IRoutingTile tile, EntityMinecart cart) {
+            return (a.evaluate(tile, cart) != 0 ? b.evaluate(tile, cart) : valueForBool(false));
```
liach: Min of these two?

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
```diff
@@ -380,8 +419,8 @@ public boolean matches(IRoutingTile tile, EntityMinecart cart) {
         }
 
         @Override
-        public boolean matches(IRoutingTile tile, EntityMinecart cart) {
-            return powered == tile.isPowered();
+        public int evaluate(IRoutingTile tile, EntityMinecart cart) {
+            return valueForBool(powered == tile.isPowered());
```
liach: return the power of the tile?

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
```diff
@@ -236,8 +254,29 @@ public OR(Condition a, Condition b) {
         }
 
         @Override
-        public boolean matches(IRoutingTile tile, EntityMinecart cart) {
-            return a.matches(tile, cart) || b.matches(tile, cart);
+        public int evaluate(IRoutingTile tile, EntityMinecart cart) {
+            int value = a.evaluate(tile, cart);
+            return (value != 0 ? value : b.evaluate(tile, cart));
```
liach: Max of two values?

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
mconwa01: I think first match, because that matches OR (last evaluated operand). The difference from the cart list case is that the order of conditions is easily controllable and the result is predictable, while the order of carts that are tried by the detector is not.

Though we could also redefine OR to be MAX, and do max value in both places. Both options seem consistent--just a matter of which is preferred.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
mconwa01: Iff we redefine OR to MAX.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
mconwa01: Again, maintaining existing behavior--this is for `Redstone=true` and `Redstone=false`. I could see adding another keyword `RedstoneLevel` or something that does that.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
mconwa01: See above.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
mconwa01: In that case, I would also clamp constants to [NO_POWER, FULL_POWER] and raise `constant.invalid` outside that range. I'm not familiar with the redstone-mod landscape--are there any that might benefit from RoutingLogic technically being able to emit a higher power than 15?

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
liach: Or the result of `&` of the two evaluated values?

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
liach: Or the result of `|` of the two evaluated values?

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
mconwa01: Ooh, bitwise could be fun. That'd mean OR (`|`) could let you check a couple independent conditions at once (anything we have a leaf condition for), as long as you could come up with a way to decode them later (e.g. limited ability with analog controller box). The only issue is that it doesn't mesh well with redstone wire signal drop, though you could always set the low bit(s) to 1s to give a little bit of distance to work with.

liach: > The only issue is that it doesn't mesh well with redstone wire signal drop, though you could always set the low bit(s) to 1s to give a little bit of distance to work with.

That is not an issue. vanilla redstone comparator sometimes outputs 1 and works fine.

mconwa01: I mean that, if you assign an independent meaning to every bit, you need to tap the signal immediately (no more than 1 block of wire), or else the signal loss will break it in annoying ways.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
CovertJaguar: I agree, use FULL_POWER and clamp to 0-15.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
CovertJaguar: Sounds complicated. Keep it simple I say, and no breaking old tables.

However, this always returns the value of b. Is that desired?

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
```diff
@@ -93,17 +93,18 @@ private EntityMinecart getRoutableCart(EntityMinecart cart) {
         return train.getLocomotive();
     }
 
-    public boolean matches(IRoutingTile tile, EntityMinecart cart) {
+    public int evaluate(IRoutingTile tile, EntityMinecart cart) {
         if (conditions == null)
-            return false;
+            return 0;
```
CovertJaguar: Use NO_POWER or define a TRUE/FALSE constant.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
```diff
@@ -133,6 +134,17 @@ private Condition parseLine(String line, Deque<Condition> stack) throws RoutingL
...
+            return new ConstantCondition(true);
+        if (line.equals("FALSE"))
+            return new ConstantCondition(false);
+        try {
+            return new ConstantCondition(Integer.parseInt(line));
+        } catch (NumberFormatException ignored) {
+            // not an integer; pass through
+        } catch (IllegalArgumentException ex) {
+            throw new RoutingLogicException("railcraft.gui.routing.logic.constant.invalid", line);
+        }
```
CovertJaguar, line 154: This feels fundamentally flawed. How do you turn a ConstantCondition into a return statement?

As written you'd have to do:

```
AND
Dest=Town
13
```

...to return 13. Which does not feel very intuitive and does not match the original issues description.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
```diff
@@ -191,7 +209,7 @@ private ParsedCondition(String keyword, boolean supportsRegex, String line) thro
         }
 
         @Override
-        public abstract boolean matches(IRoutingTile tile, EntityMinecart cart);
+        public abstract int evaluate(IRoutingTile tile, EntityMinecart cart);
```
CovertJaguar: Instead of changing matches() and spamming valueForBool() everywhere, why not just have evaluate() wrap matches()?

This also falls into the fundamental flaws I think. While its nice that Conditions can return different values, I'm not sure how desirable that is from a complexity standpoint. Like having a Condition called "InventoryFullness" that returns the comparator value for an inventory would seem to me to be shoehorning an unrelated feature into an otherwise clean Boolean logic language.

Mapping TRUE to a arbitrary return value is one thing, actually making conditions that return numbers natively I think is not a great idea.

So I'd say find some other way to make this happen rather than modifying Condition.

CovertJaguar (changes requested):
I appreciate the effort, but I fear there are some fundamental flaws in the implementation. =/

Additionally, you will have to include the documentation in the manual to get this merged.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
mconwa01: If a is false (0), it returns false; otherwise, it returns the value of b. That was my original plan, similar to how boolean short-circuit `and` and `or` work in some dynamically-typed languages, like Python and Lua.

src/main/java/mods/railcraft/common/blocks/wayobjects/RoutingLogic.java
mconwa01: That is actually what I was trying to describe in the original issue... apologies if it was unclear. Agreed that it's unintuitive, though, unless you frequently use a language where it's an idiom.

mconwa01:
How about something like this then:

```
interface Expression {
    int evaluate(...);
}
interface Condition extends Expression {
    default int evaluate(...) {
        return matches(...) ? FULL_POWER : NO_POWER;
    }
    boolean matches(...);
}
```
- Conditions otherwise as in the main branch
- Top-level elements are expressions; first nonzero is used
- New `IF <cond> <then-expr> <else-expr>` construct for hooking in values other than FULL and NO

I could add the binary conditional and bitwise operators as well, but that sounds too complicated for now. Arithmetic would raise the question of what to do if we end up outside [NO, FULL] or undefined at runtime.

CovertJaguar:
That would certainly clean up the code a bit I think.

mconwa01:
Changes pushed, documentation still to come.

Two particular notes:
- If multiple carts are next to the Routing Detector, I still take the max reading over all carts as I proposed earlier. This is just to keep it independent of cart evaluation order.
- Each operator has at least one operand that is required to be a Condition, not just any Expression. If this does not hold, it raises a RoutingLogicException, and like all such exceptions, also shows the line that failed, which is the containing operator. So if the table contains something like `IF 1 2 3` (which fails because the first operand, 1, is a general Expression and not a Condition), the user will get a message like 'Invalid Operand: "IF"'. Is this clear enough, or is it worth more juggling in `parseLine` to clarify the message?

CovertJaguar:
Looks pretty good.

liach:
@mconwa01 Can you update?

CovertJaguar (approved):

liach:
If we need to update, we can just push to his branch :hammer:

CovertJaguar:
It probably needs some testing, but the best way to get that is to probably just push it out with the routing blocks in the next release. I think I'll go ahead and merge it for now.

mconwa01:
Cursory functionality checks in current mc-1.10.2 (27e1bd1) look good. Let me know if you need me to do anything else.

PR #989, opened 2016-11-02
Fixes #988

### LINKED ISSUE #887: Command block cart will get a name of @ after quitting and joining a save game
Before quitting:

After rejoining:

### LINKED ISSUE #988: Under Cutter Cart Crashing Server
[crash-2016-11-01_18.35.30-server.txt](https://github.com/CovertJaguar/Railcraft/files/565243/crash-2016-11-01_18.35.30-server.txt)
When trying to add materials to the under cutter cart . my server crashes

liach:
Indeed I make fault sometimes.

### DIFF
--- a/src/main/java/mods/railcraft/common/util/sounds/SoundHelper.java
+++ b/src/main/java/mods/railcraft/common/util/sounds/SoundHelper.java
@@ -22,6 +22,7 @@ import net.minecraft.util.SoundEvent;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
 import net.minecraftforge.fml.common.FMLCommonHandler;
+import net.minecraftforge.fml.relauncher.ReflectionHelper;
 import net.minecraftforge.fml.relauncher.Side;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


src/main/java/mods/railcraft/common/carts/EntityCartCommand.java
```diff
@@ -44,9 +44,7 @@ protected void writeEntityToNBT(NBTTagCompound compound) {
         super.writeEntityToNBT(compound);
         // Command block logic overrode this and the custom name is
         // somewhat set to the command block's name ("@")
-        if (!getCustomNameTag().isEmpty()) {
```
CovertJaguar, line 47: hasCustomNameTag()?

CovertJaguar:
Remember when I said not to combine multiple fixes into a single commit?

src/main/java/mods/railcraft/common/carts/EntityCartCommand.java
liach: Actually, I found that even if it does not have a custom name, the cart's name would be set to the command block's default name `@`.

liach:
Wait a second

I will just fix #988

liach:
See #991 for the command block minecart issue.

PR #992, opened 2016-11-03
Add dynamic light support

### LINKED ISSUE #395: Dynamic Lights Integration
An oft requested feature is headlamps for Locomotives. The quickest and easiest way to make this happen is to hook into the existing Dynamic Lights mod.

To quote Cpt Picard, "Make it so!"

liach:
Resolves #395.
Screenshots, all tested with patched Railcraft jar:
- Without Dynamic Lights (the game does not crash or produce error):

- With Dynamic Lights, comparison of running locomotive and shutdown locomotive (shut down one does not produce light)

- With Dynamic Lights, the running one shut down and no longer emits light.

### DIFF
--- a/src/api/java/atomicstryker/dynamiclights/client/IDynamicLightSource.java
+++ b/src/api/java/atomicstryker/dynamiclights/client/IDynamicLightSource.java
@@ -1,0 +1,8 @@
+package atomicstryker.dynamiclights.client;
+
+import net.minecraft.entity.Entity;
+
+public interface IDynamicLightSource {
+    Entity getAttachmentEntity();
+    int getLightLevel();
+}
--- a/src/main/java/mods/railcraft/common/plugins/dynamiclights/DynamicLightsPlugin.java
+++ b/src/main/java/mods/railcraft/common/plugins/dynamiclights/DynamicLightsPlugin.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) CovertJaguar, 2014 http://railcraft.info
+ *
+ * This code is the property of CovertJaguar
+ * and may only be used with explicit written
+ * permission unless otherwise specified on the
+ * license page at http://railcraft.info/wiki/info:license.
+ */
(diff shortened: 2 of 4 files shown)


src/api/java/atomicstryker/dynamiclights/client/IDynamicLightSource.java
```diff
@@ -0,0 +1,8 @@
+package atomicstryker.dynamiclights.client;
+
+import net.minecraft.entity.Entity;
+
+public interface IDynamicLightSource {
```
liach, line 5: This class **will not** be included in the product jar and was not included in the patched jar in the test. It is just served as `provided` dependency.
image

CovertJaguar:
OMG you are awesome!

src/main/java/mods/railcraft/common/plugins/dynamiclights/DynamicLightsPlugin.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public void registerEntityLightSource(Class<? extends Entity> type, ToIntFunction<Entity> lightCalculator) {
+        handle.registerEntityLightSource(type, lightCalculator);
+    }
+
+    private DynamicLightsPlugin() {
+        if (FMLCommonHandler.instance().getSide().isClient()) { //only enable on client
+            try {
+                Class<?> dynamicLightsClass = Class.forName("atomicstryker.dynamiclights.client.DynamicLights");
+                Class<?> iDynamicLightSourceClass = IDynamicLightSource.class;
```
CovertJaguar, line 49: Just curious. Why is there a hard link for IDynamicLightSource but a soft link for DynamicLights?

liach: You do not need a soft link with `Class.forName()`. Either way, it will throw a `ClassNotFoundException`, handled by the `catch` block below.

Issue #998, opened 2016-11-08
Player not getting damage by electric track


tomelfring:
If the player walks on an active electric track, it doesn't get any damage. Mobs do get damage.
After checking the sourcecode I think the issue is here:
https://github.com/CovertJaguar/Railcraft/blob/mc-1.10.2/src/main/java/mods/railcraft/common/blocks/tracks/behaivor/CollisionHandler.java#L53

Suggested fix:
```
else if (entity.attackEntityFrom(RailcraftDamageSource.TRACK_ELECTRIC, 2))
    graph.removeCharge(2000);
```
If I could compile it, I would love to submit a pull request, but i'm getting the same error as #990

liach:
@thommy101 To set up you just need to run `./gradlew setupDecompWorkspace`, and then import project into IDE. You do not need to build.

tomelfring:
@liach I was in a IDE with decomp workspace, but my IDE messed up the import and used a wrong one. Reimport of the project fixed it. Thanks

PR #1009, opened 2016-11-11
Fixes issue #998

tomelfring:
As stated in issue #998, this fixes players not getting damaged by electric tracks.

Wearing Engineer's Overalls prevents damage to the user and instead damages the overalls slowly.


CovertJaguar:
Reminder to self. When merging this, refactor it to be less weird.

PR #1056, opened 2017-01-01
Fix world provider check

### LINKED ISSUE #1050: [1.10.2] Worldgeneration is limited to dimension 0.
The worldGeneration in Railcraft is limited to only generate in dimension 0. (See https://github.com/CovertJaguar/Railcraft/blob/mc-1.10.2/src/main/java/mods/railcraft/common/worldgen/GeneratorMine.java#L164 ) This creates problems with other mods, that add their own dimension, because then, RC ores won't spawn there.
It would be better to check if the biome at the specified coordinates is registered as hell or end biome in the biome dictionary to solve the problem.

liach:
Fix #1050


liach:
Not the correct solution

Issue #1046, opened 2016-12-23
[Bug] Interacting with tunnel bore in spectator crashes the game


xbjfk:
Steps to reproduce:
1: place tracks
2: place tunnel bore
3: change to spectator
4: right-click tunnel bore
5: wait

Crash Report (Railcraft is the only mod):
[crash-2016-12-23_04.35.29-server.txt](https://github.com/CovertJaguar/Railcraft/files/670278/crash-2016-12-23_04.35.29-server.txt)

liach:
Give the crash report!

liach:
How long do I need to wait? Things goes just fine:
![2017-01-01_09 56 30](https://cloud.githubusercontent.com/assets/7806504/21582356/b352dba8-d008-11e6-8299-1bf0ed36c65e.png)

AEnterprise:
@liach the crash report is original comment, he probably edited it in

liach:
Should be fixed. I tested in that pull request.

Issue #1041, opened 2016-12-17
Breaking Minecart Hopper bug


Squaredude:
Minecraft 1.10.2
Railcraft 10.0.1

Broke a Minecart Hopper and it visibly generated Hopper and Minecart entities (one of each), but the first time I did not get the Hopper back. Both times gave me the Minecart back.

However, the Minecart seems to have gotten a "Name Tag" that says "Minecart with Hopper". It's impossible to push, but it can be used as a regular Minecart (you can ride it).

If I destroy it, it gives a Minecart named "Minecart with Hopper". When I joined it with another Hopper, placing it on a rail gave me "Minecart Hopper".

Desired effect would be getting a "Minecart Hopper" with just one entity that doesn't split, correct?

snurre0:
Having the exact same bug here, same versions.

liach:
Should get closed now @CovertJaguar

PR #1057, opened 2017-01-01
Fix crash when placing tunnel bore

liach:
This fixed a potential crash. Might relate to #1046, but I am not very sure.

### DIFF
--- a/src/main/java/mods/railcraft/common/carts/EntityCartChest.java
+++ b/src/main/java/mods/railcraft/common/carts/EntityCartChest.java
@@ -47,11 +47,6 @@ public class EntityCartChest extends CartBaseContainer {
     }
 
     @Override
-    public EntityMinecart.Type getType() {
-        return EntityMinecart.Type.CHEST;
-    }
-
-    @Override
     public IBlockState getDefaultDisplayTile() {
         return Blocks.CHEST.getDefaultState().withProperty(BlockChest.FACING, EnumFacing.NORTH);
     }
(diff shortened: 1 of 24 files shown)


liach:
There are some overhauls on hopper minecarts. See #1041

src/main/java/mods/railcraft/common/carts/EntityCartChest.java
```diff
@@ -47,11 +50,6 @@ public boolean doInteract(EntityPlayer player, ItemStack stack, EnumHand hand) {
     }
 
     @Override
-    public EntityMinecart.Type getType() {
-        return EntityMinecart.Type.CHEST;
-    }
```
CovertJaguar, line 52: Why was this removed?

liach: This is intended to return `null`. The vanilla enum `EntityMinecart.Type` should be treated as deprecated under Forge, since its functionalities are covered by Forge-provided new methods. The reason not to throw an exception is because of the patch addressed [here](https://github.com/MinecraftForge/MinecraftForge/pull/3592/files#diff-7bc83f1d40c6fc8d3356e0ce7057cf9c).

CovertJaguar: This specific cart is intended to be for all intents and purposes seen as identical to the vanilla version of the chest cart. This won't change that will it?

liach:
Will fix tomorrow.

Issue #1103, opened 2017-03-08
Standardize Crafting and Breaking of Quarried, Jaded, Abyssal Stone to match Vanilla Stone Variants


Generalcamo:
Currently, Quarried Stone and Abyssal Stone mine their direct counterpart, can be smelted into their smooth counterpart, crushed into their cobble counterpart, and using smooth stone you can make bricks.

I would like to standardize this to match with stone from Minecraft and the stone variants as well. So:

Breaking Granite, Andesite, Diorite, Quarried Stone, Abyssal Stone (without Silk Touch) --> Cobble Variant (Granite, Andesite, and Diorite cobblestone will be added per #1102) 
Smelting Cobble --> Stone
4x Stone --> 4x Polished/Smooth Stone

This also removes the need for a Rock Crusher to get their cobble variants, however I'm not sure if the same should be done for the other types of cosmetic blocks, the manufactured variants.

CovertJaguar:
Sounds good.

Generalcamo:
Heavily related to #1120, so I'll be handling this alongside that for 10.2

PR #1119, opened 2017-03-21
Minor Cosmetic Recipe Changes

### LINKED ISSUE #1100: Incorporate Bone Block into Bleached Bone Recipe
Currently the Bleached Bone Block requires Bleached Clay for its creation. 1.10 however introduces Bone Blocks, which could be used in the crafting recipe instead, like how Nether Bricks were incorporated into the old recipes back in the day.

I would recommend a recipe like so:
BB*
BB*  --> 4x Bleached Bone Block
***

Where "B" is Bone Block

Generalcamo:
Closes #1100 
Part of #1103 

Made packed ice applicable for Frostbound Brick recipe
Deprecated Bleached Clay, commented out for possible usage later (as pointed out by liach)
Smelt Bone Block for Bleached Bone Block (Similar to Abyssal Stone and Quarried Stone)


src/main/java/mods/railcraft/common/blocks/aesthetics/brick/BrickTheme.java
```diff
@@ -72,6 +73,12 @@ public void initRecipes(BlockBrick block) {
                     "III",
                     'I', new ItemStack(Blocks.ICE),
                     'L', "gemLapis");
+            CraftingPlugin.addRecipe(new ItemStack(block, 8, 2),
+                    "III",
+                    "ILI",
+                    "III",
+                    'I', new ItemStack(Blocks.PACKED_ICE),
+                    'L', "gemLapis");
```
liach, line 79: `dyeBlue`?

src/main/java/mods/railcraft/common/blocks/aesthetics/brick/BrickTheme.java
```diff
@@ -52,8 +52,9 @@ public void initRecipes(BlockBrick block) {
     BLEACHEDBONE(RailcraftBlocks.BRICK_BLEACHED_BONE, MapColor.ADOBE) {
         @Override
         public void initRecipes(BlockBrick block) {
-            CraftingPlugin.addShapelessRecipe(RailcraftItems.BLEACHED_CLAY.getStack(), new ItemStack(Items.CLAY_BALL), new ItemStack(Items.DYE, 1, 15), new ItemStack(Items.DYE, 1, 15), new ItemStack(Items.DYE, 1, 15));
-            CraftingPlugin.addFurnaceRecipe(RailcraftItems.BLEACHED_CLAY.getStack(), new ItemStack(block, 1, 2), 0.3F);
+// Deprecated, possible usage in the future? ~~~GeneralCamo
+//            CraftingPlugin.addShapelessRecipe(RailcraftItems.BLEACHED_CLAY.getStack(), new ItemStack(Items.CLAY_BALL), new ItemStack(Items.DYE, 1, 15), new ItemStack(Items.DYE, 1, 15), new ItemStack(Items.DYE, 1, 15));
```
liach: You can remove these comments.

liach: Just remove these lines, recipe code is always the easiest code.

liach (changes requested):

src/main/java/mods/railcraft/common/blocks/aesthetics/brick/BrickTheme.java
Generalcamo: That's a possibility. But I want @CovertJaguar to put his opinion on that before I do it. He might have intentionally done that in the first place.

liach: So I put this part in normal comment instead of the request for change.

liach (approved):

src/main/java/mods/railcraft/common/blocks/aesthetics/brick/BrickTheme.java
CovertJaguar: Yes, Lapis only.

src/main/java/mods/railcraft/common/blocks/aesthetics/brick/BrickTheme.java
```diff
@@ -52,8 +52,7 @@ public void initRecipes(BlockBrick block) {
     BLEACHEDBONE(RailcraftBlocks.BRICK_BLEACHED_BONE, MapColor.ADOBE) {
         @Override
         public void initRecipes(BlockBrick block) {
-            CraftingPlugin.addShapelessRecipe(RailcraftItems.BLEACHED_CLAY.getStack(), new ItemStack(Items.CLAY_BALL), new ItemStack(Items.DYE, 1, 15), new ItemStack(Items.DYE, 1, 15), new ItemStack(Items.DYE, 1, 15));
-            CraftingPlugin.addFurnaceRecipe(RailcraftItems.BLEACHED_CLAY.getStack(), new ItemStack(block, 1, 2), 0.3F);
+            CraftingPlugin.addFurnaceRecipe(new ItemStack(Blocks.BONE_BLOCK), new ItemStack(block, 1, 2), 0.3F);
```
CovertJaguar: Is this not a massive nerf to the recipe? It triples the cost. I know I'd never be able to build with them at this price.

Generalcamo: To an extent. Clay is much harder to come by than Bone (as it is not renewable, and it requires aqua affinity and/or respiration for efficient mining of it). This recipe uses fully renewable resources.

If expense is a concern, doubling the amount gained from Smelting might be an option. There is no way to go back to a regular bone block, so there is no risk of duplication.

CovertJaguar: Bones only become cheap once you build some kind of mob farm, and even then, unless its an especially cheesy, tps breaking one, the output isn't that great.

Doubling the output would be fine.

Issue #423, opened 2014-12-03
Add Bronze bore head


liach:
Even through Tunnel Bore is pre-Wither, but I think that Railcraft had better include Thaumium Bore Head and enchantments.

choncho:
@liach no because Thaumiumcraft is more beginner friendly

liach:
@choncho No, I just think there'll be bore head made of void metal or something else, which can get better enchantment or special usages.

MrCalhoun:
Are there plans to overhaul the tunnel bore? I got some suggestions :)

The bore head should be able enchanted. Maybe unbreaking, efficient and some customs?

liach:
@CovertJaguar How about an invar bore head?

bookerthegeek:
Why not just dynamicaly create birr heads from any material that has an ingot? Set the harvest level to match whatever a pickaxe of the same material would have.

liach:
no... they might be only made of hard materials

liach:
Hmm will consider thaumcraft only later.

CovertJaguar:
Seems like a strange use for Invar. Nickel-Steel probably flakes too much to be useful on Bores (and too expensive) irl. But I suppose we could add one anyway.

Generalcamo:
It was probably suggested due to the occurrence of Invar tools in other mods (Substratum/Foundry and Thermal Expansion come to mind). That said, I feel the difference between the three heads is pronounced as-is.

liach:
@Generalcamo You work on this since you added metals.

Generalcamo:
There are no metals to add currently, until Thaumcraft 6 is released.

liach:
You said you wanted to add invar bores.

Generalcamo:
I never said that. I said they make no sense.

Generalcamo:
After some thought: Bronze Heads might make sense. Would be pretty much the same as Iron but with a different color.

I can do art on Void Metal and Thaumium Heads too while I am at it.

liach:
Bronze may have a lower durability, I think.

Generalcamo:
Definitely not. Bronze is equivalent to iron in strength, actually better in the ways that are relevant here.

liach:
So make them the same. Yay!

PR #1135, opened 2017-04-02
Support enchantments for bore heads

### LINKED ISSUE #813: Tunnel Bore does not respect enchanted bore heads
Vanilla behaviors permit any item that can be damaged to be enchanted with unbreaking, however the Tunnel Bore subverts the normal damage procedure and so a bore head enchanted as such will still take the full damage.

This could be implemented by simplifying [here](https://github.com/CovertJaguar/Railcraft/blob/5bd8664b24ac2f60c44143934926abdfd33e59d9/src/main/java/mods/railcraft/common/carts/EntityTunnelBore.java#L800-L802): `if (head.attemptDamageItem(1, rand)) setInventorySlotContents(0, null);`

Somewhat related, not sure if I should make a separate issue for it however; it would be nice if at the same place, the Bore would attempt to replace its head from the attached minecarts inside the same conditional, such that the bore can have a stock of the fuel, tracks, and heads it needs to make it to a destination, barring it running into any other trouble.

liach:
Supports efficiency, unbreaking, and fortune.
Attacking mobs harms durability, 1 for each attack for each mob
Fixes #813, ref #423


Generalcamo:
Could this support Silk Touch too? @liach

liach:
Good point, although I don't have much time. Feel free to push.

liach:
@Generalcamo After looking at the code, I found `getDrops` already supports silk touch. So this is fine.

src/main/java/mods/railcraft/common/carts/EntityTunnelBore.java
```diff
@@ -454,9 +456,10 @@ public void onUpdate() {
 
                     if (boreLayer) {
                         boolean bored = boreLayer(targetPos, dir);
-                        if (bored)
-                            setDelay(LAYER_DELAY);
-                        else {
+                        if (bored) {
+                            int efficiency = EnchantmentHelper.getEnchantmentLevel(Enchantments.EFFICIENCY, getStackInSlot(0));
+                            setDelay(LAYER_DELAY / (int)(efficiency * .2 + 1));
```
CovertJaguar: Does this still give enough time for gravity affected blocks like sand to fall before moving? That's the entire reason this delay even exists.

It would probably make more sense to move this to getLayerHardness() where the RailcraftConfig.boreMiningSpeedMultiplier() is applied.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/carts/EntityTunnelBore.java
liach: hmm, did not find a better place at first.

liach:
@CovertJaguar Fixed.

src/main/java/mods/railcraft/common/carts/EntityTunnelBore.java
```diff
@@ -857,6 +865,7 @@ protected float getLayerHardness(BlockPos targetPos, BlockRailBase.EnumRailDirec
             IBoreHead head = (IBoreHead) boreSlot.getItem();
             float dig = 2f - head.getDigModifier();
             hardness *= dig;
+            hardness /= (1 + EnchantmentHelper.getEnchantmentLevel(Enchantments.EFFICIENCY, boreSlot));
```
CovertJaguar: I cannot match this algorithm with the numbers you posted.

Level 1 cuts the time in half and you get diminishing returns from there.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/carts/EntityTunnelBore.java
liach: This algorithm is changed to match vanilla efficiency tool behavior. I will change my description above.

CovertJaguar: "If a proper tool is used, the tool speed multiplier is further increased by Efficiency. If the level of Efficiency is not 0, then the level squared plus 1 is added to the tool speed. For example, Efficiency I adds 2 to the value, while Efficiency V adds 26."

That doesn't sound much like what you implemented.

liach:
Wait, the CLA license check seems to break @CovertJaguar

src/main/java/mods/railcraft/common/carts/EntityTunnelBore.java
liach: Will fix.

liach: Btw do bores only mine pickaxe and shovel stuff or everything?

CovertJaguar: pickaxe shovel and axe I think, the code is then there somewhere with the tool types it looks at

liach:
@CovertJaguar Fixed.

src/main/java/mods/railcraft/common/carts/EntityTunnelBore.java
```diff
@@ -857,15 +857,22 @@ private boolean canMineBlock(BlockPos targetPos, IBlockState existingState) {
     }
 
     protected float getLayerHardness(BlockPos targetPos, BlockRailBase.EnumRailDirection dir) {
-        float hardness = layerAction(targetPos, dir, 0F, this::getBlockHardness, (s, r) -> s + r);
+        float hardness = layerAction(targetPos, dir, 0F, (pos, direction) -> {
+            float f = getBlockHardness(pos, direction);
+            ItemStack head = getStackInSlot(0);
+            if (canHeadHarvestBlock(head, worldObj.getBlockState(pos))) {
+                int i = EnchantmentHelper.getEnchantmentLevel(Enchantments.EFFICIENCY, head);
+                f /= i * i + 1;
```
CovertJaguar: What is this? This is even worse that before! And its totally not necessary to insert it into the summation.

Level 1: 2x faster
Level 2: 5x faster
Level 3: 10x faster
Level 4: 17x faster
Level 5: 26x faster

How Tools work:
```
strVsBlock = wood 2 || stone 4 || iron 6 || diamond 8
digSpeed = strVsBlock + enchant * enchant + 1
blockDamagePerTick =  digSpeed  / hardness / 30F

tick() {
    blockDamage += blockDamagePerTick 
    if (blockDamage > 1) destroyBlock()
}
```

A Diamond Pick with Level 5 Efficiency is only 12x faster than a Wooden Pick with Level 0.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/carts/EntityTunnelBore.java
liach: But we don't have something like `strVsBlock`

CovertJaguar: Nope, that means we are going to need a brand new algorithm, one that produces similar numbers to the enchantments affect on tools but that is tailored to our situation.

liach: So can we just use the original algorithm in the comment?

CovertJaguar: This one?
```
hardness /= (1 + EnchantmentHelper.getEnchantmentLevel(Enchantments.EFFICIENCY, boreSlot));
```
No, it has very non-linear characteristics, making Level 5 nearly worthless compared to Level 1.

liach: How about divide by `(efficiency * .2 + 1)`?

CovertJaguar: It's still diminishing returns and non-linear, and it caps out at 2x faster. That's probably not sufficient. Maybe a scaled version of f /= i * i + 1 would work. Something like f /= i * i * 0.2 + 1 maybe.

That caps out about 5x faster, and is more linear.

liach: Can you edit on the branch on your own? I am not really sure what speed of mining do you want.

CovertJaguar (approved):

liach:
Squash merge?

PR #1140, opened 2017-04-04
Added Nickel[WIP]

Generalcamo:
For review purposes only.

Currently broken, unknown reasoning


Generalcamo:
Turns out that there are too many ores. Going to need to split them, so this is going to take longer than previously thought.

Generalcamo:
Closed as this requires more work internally.

CovertJaguar:
Instead of doing loot, I'll make splitting the blocks my next project.

PR #1160, opened 2017-04-21
Reverse mode for throttle track

### LINKED ISSUE #1159: Add a reverse option to the throttle track.
Selfy explanatory.

This one's on me.

Cream-Tea:
I have added a reverse mode functionality to the throttle track, since it doesn't seem to be a feature at the moment, yet the old limiter track used to have a reverse mode.

Closes #1159


src/main/java/mods/railcraft/common/blocks/tracks/outfitted/kits/TrackKitThrottle.java
```diff
@@ -40,14 +41,24 @@ public TrackKits getTrackKitContainer() {
...
-        if (isPowered())
-            state += LocoSpeed.VALUES.length;
+        if (getReverse())
+        	state = 14;
+        if (isPowered()) {
+            if (getReverse())
+            	state += 1;
+            else
+            	state += 7;
+    	}
```
CovertJaguar: You can simplify this by reordering the states.

src/main/resources/assets/railcraft/blockstates/tracks/outfitted/kit/throttle.json
```diff
@@ -86,6 +86,17 @@
...
+        "textures": {
+          "layer0": "railcraft:blocks/tracks/outfitted/kit/limiter_r",
+        }
+      }
+      "15": {
+        "textures": {
+          "layer0": "railcraft:blocks/tracks/outfitted/kit/limiter_r",
+          "layer1": "railcraft:blocks/tracks/outfitted/kit/throttle_overlay_active"
+        }
+      }
```
CovertJaguar: The six "limiter" texture that never got renamed are intended for reverse mode. If you wanted to add more reverse speeds, you could do it by reintegrating these textures.

CovertJaguar (commented):
Just so everyone is on the same page, perhaps some history in order. The track was intended originally to have seven speed modes, three of which were to be reverse. And this was true, until I changed locomotives to handle speed and direction separately. When I did this, the Throttle track lost its reverse capability because the LocoSpeed enum was now only four elements long instead of seven/eight. So if we were to do a proper pass at this, we should clean up those unused reverse "limiter" textures, either rename them to bring them in line with the existing textures and use them, or just delete them entirely and remove the entries from the blockstate JSON.

version/gradle.properties
```diff
@@ -1,11 +1,11 @@
...
 version_forestry=[5.1.15.295,)
-version_industrialcraft=[2.6.11-ex110,2.7.0-ex111)
+version_industrialcraft=2.6.188-ex110
 version_project_major=10
 version_project_series=1
 version_project_revision=0
-version_project_build=0
-version_project_qualifier=none
+version_project_build=3.1
+version_project_qualifier=beta
```
CovertJaguar: You'll need to revert these two lines.

version/gradle.properties
```diff
@@ -1,11 +1,11 @@
-#Thu, 20 Apr 2017 11:30:12 -0700
+#13:17 21/04/2017
 version_minecraft=1.10.2
 version_forge=12.18.3.2239
 version_mcp=snapshot_20160826
 version_forestry=[5.1.15.295,)
-version_industrialcraft=[2.6.11-ex110,2.7.0-ex111)
+version_industrialcraft=2.6.188-ex110
```
CovertJaguar: I could have sworn the existing bounds should have resolved this issue.

CovertJaguar (changes requested):

version/gradle.properties
Cream-Tea: Alright I'll revert them, I wasn't really too sure about what they were supposed to be.

CovertJaguar (approved):

Issue #662, opened 2015-07-19
[Request] Move recipes to Init instead of PreInit


BluSunrize:
Basically, people keep filing error reports that Immersive Engineering's creosote bottles+buckets do not work to create wooden ties. The reason for that is that IE registers it's FluidContainers in PreInit, but not before Railcraft has already registered it's recipes for wooden ties. @malte0811 on my issue tracker has already traced this:

https://github.com/BluSunrize/ImmersiveEngineering/issues/101

The obvious solution would be to move the registration of those recipes to the init phase, assuming that this does mess up other parts of the code (Haven't checked that yet ^^"). A different, probably less breakage inducing way (but not as pretty) would be to not use defineRecipes() in ItemTie and instead register the recipes manually in the init phase.

JonathanMH:
yes please +1 :)

smbarbour:
Recipes should not be done in preInit.  Recipes should be done in init or postInit so that all mods have a chance to register their items before recipes are set.

BluSunrize:
Thank you. Now if only Covert or one of the Railcraft maintainers (provided there are some) would aid with this >_>

vpontin:
related to my issue regrading CraftGuide? #679

CovertJaguar:
Hmm...yes fair point.

PR #1177, opened 2017-04-26
Reorganized to make Railcraft more modular

### LINKED ISSUE #1167: (CRASH) "Ticking Entity" after trading with Railcraft Villager
Minecraft 1.10.2 / Railcraft 10,1.1 / Forge 2281

User was performing a trade with a Railcraft villager. Mid-trade, server crashes due to a "Ticking Entity" error. [https://github.com/Vaygrim/CoopLife/issues/7](https://github.com/Vaygrim/CoopLife/issues/7)

First time it has happened, doesn't _seem_ common?

liach:
Now every railcraft object has a module.
When the module is disabled, the object is also disabled. Moved recipe registering to init stage to avoid conflicts with mods like immersive engineering (reported before). Also avoided unorganized loading of objects because of unorganized recipes. Removed silly code in `ItemLocoElectric.java`. Added more fancy info of item register failure. Finally, Fixes #1167, tested with signal module disabled. Before the game crashes when the villager unlocks tier 3 trade, and now trades with signal items simply disappear. Same stack trace in the end, although neat mod was not involved.


CovertJaguar:
You realize items are intentionally intended to have lazy initialization right? They are only enabled if something uses them.

How is this an improvement on that?

CovertJaguar (changes requested):
Make a case for why this should be accepted.

liach:
Main ideas:
1. Keeping a reference to the modules allows to find out specific issues when certain modules are disabled better.
2. Loading items and blocks during recipe definition is a bad behavior. They will be loaded in random order. Worst of all is that a recipe may load unwanted blocks/items from a disabled module.
3. Moving recipes from preinit to init is a necessity. See #662.

CovertJaguar:
1. Fair point

2. Items don't have modules... as I said, they are loaded as needed. This was by design. It helps to keep you from breaking your crafting tree. And random order registration is an issue how?

3. I don't see how that issue is relevant as it refers to a specific recipe that was scanning the fluid registry for items.

liach:
Well, in the module classes, items are included by modules. 

> Items don't have modules... as I said, they are loaded as needed. This was by design. It helps to keep you from breaking your crafting tree.

This is what caused #1167: You expect items not to be modular but often times, when you find them in need, it is too late to load.

The last point is about most recipes, not only fluids. That is general design, I will ask @sfPlayer1.

src/main/java/mods/railcraft/common/modules/ModuleCore.java
```diff
@@ -131,8 +131,10 @@ public void construction() {
                         RailcraftItems.MAG_GLASS,
                         RailcraftItems.GOGGLES,
                         RailcraftItems.OVERALLS,
-                        RailcraftItems.NUGGET,
-                        RailcraftItems.NOTEPAD
+                        RailcraftItems.NOTEPAD,
+
+                        RailcraftItems.RAIL,
+                        RailcraftItems.TIE
```
CovertJaguar, line 137: Would track module be better?

src/main/java/mods/railcraft/common/worldgen/VillagerTrades.java
```diff
@@ -49,12 +50,14 @@ public static void define(VillagerRegistry.VillagerCareer career) {
...
         career.addTrade(1, new GenericTrade(offer(Blocks.RAIL, 30, 34), offer(Items.EMERALD, 2, 3)));
 
-        for (Map.Entry<String, TrackKit> track : TrackRegistry.TRACK_KIT.getVariants().entrySet()) {
-            TrackKit kit = track.getValue();
-            if (kit != TrackRegistry.getMissingTrackKit() && kit.isEnabled()) {
-                career.addTrade(2, new GenericTrade(offer(kit.getTrackKitItem(), 2), offer(Items.EMERALD, 1, 3)));
+//        if (RailcraftBlocks.TRACK_OUTFITTED.isEnabled()) {
+            for (Map.Entry<String, TrackKit> track : TrackRegistry.TRACK_KIT.getVariants().entrySet()) {
+                TrackKit kit = track.getValue();
+                if (kit != TrackRegistry.getMissingTrackKit() && kit.isEnabled()) {
```
CovertJaguar: Probably check visible instead of missing.

CovertJaguar (changes requested):

CovertJaguar:
Ok, I did a thorough review and its not as crazy as it first looked. It'll probably break everything though, so I'll have to sit on it until I'm done with hotfixes for 10.1.0.

liach:
The trade on track kit is currently useless, we probably need some method to poll a random one like how librarians trade ench books.

Generalcamo:
Librarians use the Enchantment code to get a random enchant of x levels, which includes treasure enchants. That would be very difficult to implement.

liach:
@Generalcamo That support is within the registry of enchantment. If we need to do the same, we might seek to tweak track registry to make this more convenient (not necessary though)

src/main/java/mods/railcraft/common/modules/ModuleCore.java
liach: Not really, this feature has always been in core.

src/main/java/mods/railcraft/common/worldgen/VillagerTrades.java
liach: #1181

liach:
@CovertJaguar Can you review?

CovertJaguar:
Oh right, I forgot about this.

CovertJaguar (approved):

PR #1178, opened 2017-04-26
Fix supplying null entries into the creative list

### LINKED ISSUE #1171: Creative Mode crash with certain blocks disabled
Railcraft will crash using the Creative Search tab (not the Railcraft tab) with certain blocks disabled.  From the testing I did, these blocks were the ones at fault:

    B:ore_dark_diamond=false
    B:ore_dark_emerald=false
    B:ore_dark_lapis=false
    B:ore_saltpeter=false
    B:ore_sulfur=false

Other ores can still be disabled safely.

May be related to #1162.

edit: Ah, almost forgot a crash log: https://pastebin.com/su0dktz2

liach:
Fixed #1171 after the second commit. Before the game instantly crashes when you click that search tab. Now there is no effect, although loot table warns for missing items when the signal module is disabled (does not crash game)

### DIFF
--- a/src/main/java/mods/railcraft/common/util/misc/Predicates.java
+++ b/src/main/java/mods/railcraft/common/util/misc/Predicates.java
@@ -15,6 +15,7 @@ import mods.railcraft.common.util.collections.StackKey;
 import net.minecraft.item.ItemStack;
 
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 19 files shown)


src/main/java/mods/railcraft/common/util/misc/Predicates.java
```diff
@@ -24,7 +25,13 @@
  *
  * @author CovertJaguar <http://www.railcraft.info>
  */
+@SuppressWarnings({"unchecked", "rawtypes"})
 public class Predicates {
+
+    private static final Predicate NON_NULL = Objects::nonNull;
+    private static final Predicate ALWAYS_FALSE = t -> false;
+    private static final Predicate ALWAYS_TRUE = t -> true;
```
CovertJaguar, line 33: You are breaking the generics, keep your PRs focused. Ignoring the warnings is stupid.

CovertJaguar (changes requested):
Removes the changes to Predicates, and it looks good.

src/main/java/mods/railcraft/common/util/misc/Predicates.java
liach: It does not break. You should use `rawtypes`, just like `java.util.Collections.emptyList()` returns a generic list casted from a rawtype list. [Link to the source](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/Collections.java#l4458)

CovertJaguar: Why did you even make the change?

liach: To prevent creating extra objects.

CovertJaguar (approved):

PR #1186, opened 2017-04-30
General improvements to particles

Generalcamo:
Basically in this pull I changed the properties of the sound for Zaps to support pitch and volume bending, and also added a sound for Firestone Ore.

In addition, I added smoke particles to Firestone Ore's fire fountain.

### DIFF
--- a/src/main/java/mods/railcraft/client/particles/ParticleFireSpark.java
+++ b/src/main/java/mods/railcraft/client/particles/ParticleFireSpark.java
@@ -11,6 +11,7 @@ package mods.railcraft.client.particles;
 
 import net.minecraft.client.renderer.VertexBuffer;
 import net.minecraft.entity.Entity;
+import net.minecraft.util.EnumParticleTypes;
 import net.minecraft.util.math.Vec3d;
 import net.minecraft.world.World;
 import net.minecraftforge.fml.relauncher.Side;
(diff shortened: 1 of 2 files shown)


src/main/java/mods/railcraft/client/particles/ParticleFireSpark.java
```diff
@@ -100,6 +101,16 @@ public void onUpdate() {
...
 
+        /**
+         * Called to spawn smoke particles with the entity
+         */
+        float f = (float)this.particleAge / (float)this.particleMaxAge;
+
+        if (this.rand.nextFloat() > f)
+        {
+            this.worldObj.spawnParticle(EnumParticleTypes.SMOKE_NORMAL, this.posX, this.posY, this.posZ, this.motionX, this.motionY, this.motionZ, 0);
+        }
```
liach, line 112: 😞

CovertJaguar: I did think this was a little strange, I'll need to check what it does to the recharge animation too.

liach: code style and javadoc!

CovertJaguar: oh that...

PR #1188, opened 2017-05-01
Some fixes and improvements to Backpacks

Generalcamo:
This adds some new items to the Trackman's Backpack, but also splits it into the existing Trackman's Backpack for tracks and carts, and the Signalman's backpack for Signals and Detectors.

In addition, JEI descriptions were added for all Railcraft Backpacks.


src/main/java/mods/railcraft/common/plugins/forestry/SignalmanBackpack.java
```diff
@@ -0,0 +1,78 @@
...
+        for (ResourceLocation id : Block.REGISTRY.getKeys()) {
+            Block block = Block.REGISTRY.getObject(id);
+            if (block == null) continue;
+            if (TrackTools.isRailBlock(block))
+                add(block);
+        }
+
+        for (ResourceLocation id : Item.REGISTRY.getKeys()) {
+            Item item = Item.REGISTRY.getObject(id);
+            if (item instanceof ItemMinecart || item instanceof IMinecartItem)
```
CovertJaguar: Do we want tracks and minecarts here?

PR #1208, opened 2017-05-16
Refactors for track registry

liach:
Cleanup after api pull.
Seems it cannot grab the correct submodule, so it fails to build.
See https://github.com/CovertJaguar/Railcraft-API/pull/18


liach:
As long as you merge the api one first and rebuild, this will pass.

src/main/java/mods/railcraft/common/util/misc/Game.java
```diff
@@ -33,17 +32,13 @@
...
+            obfuscated = MinecraftServer.class.getDeclaredField("worldServers") == null;
+        } catch (NoSuchFieldException absence) {
+            obfuscated = true;
         }
-        try {
-            worldObjFound = Entity.class.getDeclaredField("worldObj") != null;
-        } catch (NoSuchFieldException | SecurityException ignored) {
-        }
-        OBFUSCATED = !worldFound && !worldObjFound;
+        OBFUSCATED = obfuscated;
```
CovertJaguar: Why are we changing this shit again? I just got done testing it in all possible environments. -.-

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/util/misc/Game.java
liach: Cuz this is done after merging. You should consider this as a merge conflict, and I will fix it too,

CovertJaguar (approved):

liach:
Let me fix submodules.

liach:
Guess I will merge this one first

PR #1211, opened 2017-05-17
Fixed #1004, do not dump if there is block below

### LINKED ISSUE #1004: Dumping Track dumps mobs into the ground
And then they suffocate.
 rc 10.0.0.0
forge x2122

liach:


src/main/java/mods/railcraft/common/blocks/tracks/outfitted/kits/TrackKitDumping.java
```diff
@@ -34,7 +34,7 @@ public TrackKits getTrackKitContainer() {
 
     @Override
     public void onMinecartPass(EntityMinecart cart) {
-        if (!isPowered()) {
+        if (!isPowered() && !theWorldAsserted().getBlockState(getPos().down()).getBlock().isVisuallyOpaque()) {
```
CovertJaguar: I'd suggest looking at the block below this as well, since it drops the entity two blocks down.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/blocks/tracks/outfitted/kits/TrackKitDumping.java
liach: That block does not hurt to players or villagers, etc. They cannot suffocate when their feet is in block

Generalcamo: It still gets them stuck though, which is unintended behavior.

liach: I don't think they will stuck. Anyways I will adjust it

CovertJaguar (approved):

PR #1227, opened 2017-05-27
Fixed launch crash

liach:
Also adjusted rarity to make smack less often

Hotfix @Generalcamo


Generalcamo (approved):

CovertJaguar:
Ah I hadn't actually tested this yet, ran out of time.

PR #1243, opened 2017-06-06
Improved buffer stop track hitbox detection

### LINKED ISSUE #1221: Reduce Buffer stop track hitbox
When I try to boost a minecart next on the buffer, the crowbar targets the buffer even if I look at the minecart. I suggest reducing the size of the stopping buffer. Pinging @CovertJaguar for opinions.

liach:
Tested both track kits, they are fine.
Fixes #1221


liach:
Looking at the hitbox
Not looking at the hitbox

CovertJaguar (approved):

PR #1244, opened 2017-06-07
Colorizable smokers

### LINKED ISSUE #656: [Suggestion] Colourizable smokers
How about adding an option to the smokers that would enable changing the smoke color? Like, right-clicking with a dye, or having a dye somewhere in the crafting recipe? Not all smoke is black :)

liach:
Closes #656
Smokers now can be colored by using any ore dictionary dye on the smoker
with main hand while not sneaking. It consumes one dye. Adds support for
comparator output for smoker color.
There are some minor fixes needed for inventories, too.


liach:
Will get a screenshot soon.

liach:
Smokers and redstone output

Generalcamo:
I'm not entirely sure why a smoker needs to output comparator redstone signals based on color. But I suppose you could say the same for Jukeboxes and the like.

src/main/java/mods/railcraft/common/util/inventory/InvTools.java
```diff
@@ -78,6 +78,34 @@ public static ItemStack emptyStack() {
         return null;
     }
 
+    /**
+     * This method changes the original stack!
+     */
+    @Nullable
+    public static ItemStack decSize(@Nullable ItemStack stack) {
```
CovertJaguar: This is basically an incomplete version of depleteItem(ItemStack).

liach: For deplete one, does it mutate the stack itself? I don't understand it very well so I rewrote a size decrease function that always changes the stack in the argument.

CovertJaguar: Mutating the stack is a bad idea generally. I've gotten tons of flack for doing that in the past from other modders who use object identity to detect changes to stacks.

liach: Most of the uses of `ItemStack#stackSize` field are direct mutations, so it is not that bad an idea. In this case, this stack is in player's hand, consuming one after dying the smoker.

CovertJaguar: But I guess it might have its uses.

CovertJaguar: However in the case of the Smoker, it should definitely use depleteItem() for the container support.

liach:
@CovertJaguar Build was killed by missing `BlockChargeBattery.java`, not caused by pr contents.

src/main/java/mods/railcraft/common/blocks/machine/equipment/TileSmoker.java
```diff
@@ -76,25 +88,55 @@ public boolean isSideSolid(EnumFacing side) {
...
+    }
+
+    @Override
+    public boolean blockActivated(EntityPlayer player, EnumHand hand, @Nullable ItemStack heldItem, EnumFacing side, float hitX, float hitY,
+        float hitZ) {
+        if (super.blockActivated(player, hand, heldItem, side, hitX, hitY, hitZ))
+            return true;
+        if (player.isSneaking())
+            return false;
+        if (heldItem == null || hand == EnumHand.OFF_HAND)
```
Generalcamo: Use InvTools.isEmpty(heldItem)

src/main/java/mods/railcraft/common/blocks/machine/equipment/TileSmoker.java
```diff
@@ -67,6 +77,8 @@ public void onNeighborBlockChange(IBlockState state, Block block) {
         sendUpdateToClient();
     }
 
+
+
```
Generalcamo: Seems like a large amount of spacing added here.

Generalcamo (dismissed):

Generalcamo (approved):
Wait for CJ's approval, but I see no issues.

CovertJaguar (approved):

PR #1252, opened 2017-06-11
Fix destruction thing

### LINKED ISSUE #1250: Destruction enchantment don't drop
It does not drop blocks destroyed by the enchantment.

liach:
Tested and confirmed this fixes #1250. Also tried to improve code for compatibility.

P.S. I was not really sure what does `Block#removedByPlayer` do.

### DIFF
--- a/src/main/java/mods/railcraft/common/plugins/forge/WorldPlugin.java
+++ b/src/main/java/mods/railcraft/common/plugins/forge/WorldPlugin.java
@@ -136,6 +136,7 @@ public class WorldPlugin {
         if (player == null)
             player = RailcraftFakePlayer.get((WorldServer) world, pos);
         IBlockState state = getBlockState(world, pos);
+        TileEntity te = getBlockTile(world, pos);
 
         if (MinecraftForge.EVENT_BUS.post(new BlockEvent.BreakEvent(world, pos, state, player)))
             return false;
@@ -144,7 +145,7 @@ public class WorldPlugin {
             return false;
 
         if (dropBlock) {
-            state.getBlock().harvestBlock(world, player, pos, state, getBlockTile(world, pos), player.getHeldItem(player.getActiveHand()));
+            state.getBlock().harvestBlock(world, player, pos, state, te, player.getHeldItem(player.getActiveHand()));
         }
         return true;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/mods/railcraft/common/plugins/forge/WorldPlugin.java
```diff
@@ -140,12 +140,11 @@ public static boolean playerRemoveBlock(World world, BlockPos pos, @Nullable Ent
...
             return false;
 
-        if (state.getBlock().removedByPlayer(state, world, pos, player, dropBlock))
+        if (!state.getBlock().removedByPlayer(state, world, pos, player, dropBlock))
             return false;
 
         if (dropBlock) {
-            List<ItemStack> drops = state.getBlock().getDrops(world, pos, state, 0);
-            InvTools.dropItems(drops, world, pos);
+            state.getBlock().harvestBlock(world, player, pos, state, getBlockTile(world, pos), player.getHeldItem(player.getActiveHand()));
```
CovertJaguar: I don't know what you are trying to do here, but the tile no longer exists at this point, so calling getBlockTile(world, pos) will return null always.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/plugins/forge/WorldPlugin.java
liach: From the javadoc of `Block#removedByPlayer`:
```
    * @param willHarvest True if Block.harvestBlock will be called after this, if the return in true.
    *        Can be useful to delay the destruction of tile entities till after harvestBlock
```
The method I call above allows blocks to guarantee that the tile would not get destroyed until `#harvestBlock` is called.

CovertJaguar: And what of it? Have you looked at the actual code? It doesn't obey that.

liach: That simply says if you want to keep the tile you can keep it in destroyBlock and remove it in harvestBlock, or remove it directly when it does not drop or when the drop does not depend on the tile entity. So my code is fine.

liach: Moreover, in `harvestBlock` method the tile entity argument is nullable.

CovertJaguar: "Is nullable" doesn't mean it should be null at all times. Doesn't the vanilla code cache the tile entity before calling removedByPlayer?

liach: Hmm I will change then. Wait for me to work on the other features first.

CovertJaguar (approved):

PR #1253, opened 2017-06-11
Adjusted link message to be above hotbar

### LINKED ISSUE #1251: [Suggestion] Put linking message to the area for "Press LShift to dismount"
This may make things neater. Will decide to implement or not per @CovertJaguar's opinion,

liach:
Closes #1251

### DIFF
--- a/src/main/java/mods/railcraft/common/plugins/forge/ChatPlugin.java
+++ b/src/main/java/mods/railcraft/common/plugins/forge/ChatPlugin.java
@@ -12,6 +12,8 @@ package mods.railcraft.common.plugins.forge;
 import com.mojang.authlib.GameProfile;
 import mods.railcraft.common.util.misc.Game;
 import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.entity.player.EntityPlayerMP;
+import net.minecraft.network.play.server.SPacketChat;
 import net.minecraft.util.text.ITextComponent;
 import net.minecraft.util.text.TextComponentString;
 import net.minecraft.util.text.TextComponentTranslation;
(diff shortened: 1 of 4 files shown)


liach:
Should I change outfitted track placement message to be like these too?

Generalcamo:
This adds basically consistency with the changes in 1.11. I approve.

My opinion is that everything that isn't a chat command should appear in the hotbar.

liach:
So I think we can put the message for track placement and goggles auras here too, but not those for commands.

Moreover, I will leave those more informatic ones as it is, such as messages from magnifying glass. Track placement, cart linking, and goggles auras are easy to spam logs. (You can see goggles aura in item tooltips too)

src/main/java/mods/railcraft/common/plugins/forge/ChatPlugin.java
```diff
@@ -65,6 +67,21 @@ public static void sendLocalizedChatFromServer(@Nullable EntityPlayer player, St
...
+        if (player instanceof EntityPlayerMP && Game.isHost(player.worldObj)) {
+            EntityPlayerMP target = (EntityPlayerMP) player;
+            for (int i = 0; i < args.length; i++) {
+                if (args[i] instanceof String) {
+                    args[i] = translateMessage((String) args[i]);
+                } else if (args[i] instanceof GameProfile) {
+                    String username = ((GameProfile) args[i]).getName();
+                    args[i] = username != null ? username : "[unknown]";
+                }
+            }
```
CovertJaguar: Copy paste code block? Split it out into a function please.

CovertJaguar (changes requested):

CovertJaguar (approved):

PR #1260, opened 2017-06-20
Defined a charge terminal in charge trap recipe

### LINKED ISSUE #1155: Add a real recipe for the Charge Trap
The Charge Trap is currently using a temporary recipe. It needs a better recipe.

liach:
Yes - It becomes more expensive
Fixes #1155


liach:
@CovertJaguar This does not have a battery since it is not defined to have one; it just has a charge terminal for releasing currents.

CovertJaguar (approved):

PR #1261, opened 2017-06-20
Various worldspike fixes

### LINKED ISSUE #847: Attempting to complete pairing with an anchor/sentinel which no longer exists does not cause an error
This is probably a low priority bug, given that there are two workarounds, but I thought it worth posting anyway (and I didn't see an existing bug report).

Basically, if you're trying to pair an anchor with a sentinel, and you destroy the anchor/sentinel that you BEGAN the pairing process with, you cannot complete or cancel the pairing process. Once this occurs, you cannot begin pairing on a new anchor/sentinel\* (*see workarounds).

In my opinion, if you try to complete pairing on an anchor/sentinel which no longer exists, you should receive an error to the effect of "the pair's origin no longer exists," and pairing should be cancelled allowing you to start the pairing process again. 

Minecraft: 1.7.10
Forge: 10.13.4.1614
Railcraft: 9.12.2

Here's a video:
https://youtu.be/j62CTe4fJUk

Reproduction steps:
1. Place an anchor and a sentinel anywhere in the world
2. While holding a crowbar, right click the sentinel to begin pairing
3. Destroy the sentinel
4. While holding a crowbar, right click the anchor -- normally this would produce some kind of message, but if a sentinel doesn't exist at the sentinel's original set coordinates, nothing happens
5. Create another sentinel (at a different set of coordinates)
6. While holding a crowbar, right click the sentinel to begin pairing -- in this case, you receive no error message, and you have no way\* (see workarounds) to pair the new sentinel with the anchor

Workaround:
- Logout of the server, or end single player game. Log back in, and re-pair; OR
- Place an anchor/sentinel at the original set of coordinates, complete pairing (either by pairing or cancelling) and destroy the anchor/sentinel

### LINKED ISSUE #1258: Worldspike Point xrays
image

liach:
Fixes #1258
Fixes #847 - Always clean up when failing to pair. Now missing ones will
be said to be "invalid"
Also fixed localized name of worldspike points.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/machine/worldspike/BlockWorldspikePoint.java
+++ b/src/main/java/mods/railcraft/common/blocks/machine/worldspike/BlockWorldspikePoint.java
@@ -54,21 +54,22 @@ public class BlockWorldspikePoint extends BlockRailcraft {
             if (crowbar.canWhack(player, hand, heldItem, pos)) {
                 WorldCoordinate ourCoord = new WorldCoordinate(world.provider.getDimension(), pos);
                 WorldCoordinate target = TileWorldspike.getTarget(player);
-                if (target == null)
-                    TileWorldspike.setTarget(ourCoord, player, getUnlocalizedName());
-                else if (world.provider.getDimension() != target.getDim())
-                    ChatPlugin.sendLocalizedChatFromServer(player, "gui.railcraft.worldspike.pair.fail.dimension", getUnlocalizedName());
-                else if (ourCoord.equals(target)) {
-                    TileWorldspike.removeTarget(player);
-                    ChatPlugin.sendLocalizedChatFromServer(player, "gui.railcraft.worldspike.pair.cancel", getUnlocalizedName());
+                if (target == null) {
+                    TileWorldspike.setTarget(ourCoord, player, getLocalizedName());
                 } else {
-                    if (TileWorldspike.isTargetLoaded(player, target, getUnlocalizedName())) {
+                    if (world.provider.getDimension() != target.getDim()) {
+                        ChatPlugin.sendLocalizedChatFromServer(player, "gui.railcraft.worldspike.pair.fail.dimension", getLocalizedName());
+                    } else if (ourCoord.equals(target)) {
+                        ChatPlugin.sendLocalizedChatFromServer(player, "gui.railcraft.worldspike.pair.cancel", getLocalizedName());
+                    } else if (TileWorldspike.isTargetLoaded(player, target, getLocalizedName())) {
                         TileEntity tile = WorldPlugin.getBlockTile(world, target.getPos());
-                        if (tile instanceof TileWorldspike)
+                        if (tile instanceof TileWorldspike) {
                             ((TileWorldspike) tile).setPoint(player, ourCoord);
-                        else if (tile != null)
-                            ChatPlugin.sendLocalizedChatFromServer(player, "gui.railcraft.worldspike.pair.fail.invalid", getUnlocalizedName());
+                        } else {
+                            ChatPlugin.sendLocalizedChatFromServer(player, "gui.railcraft.worldspike.pair.fail.invalid", getLocalizedName());
+                        }
                     }
+                    TileWorldspike.removeTarget(player);
                 }
                 crowbar.onWhack(player, hand, heldItem, pos);
                 return true;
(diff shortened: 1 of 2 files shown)


CovertJaguar (changes requested):
There is Style settings file in extras. Use it.

src/main/java/mods/railcraft/common/blocks/machine/worldspike/BlockWorldspikePoint.java
```diff
@@ -84,20 +86,29 @@ public float getBlockHardness(IBlockState state, World worldIn, BlockPos pos) {
...
+                "ogo",
+                'g', "ingotGold",
+                'p', Items.ENDER_PEARL,
+                'o', new ItemStack(Blocks.OBSIDIAN));
         }
     }
+
+    @Override
+    public String getUnlocalizedName() {
+        return super.getUnlocalizedName() + ".name";
```
CovertJaguar: What is the point of this?

This is basically the same as getLocalizedName().

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/blocks/machine/worldspike/BlockWorldspikePoint.java
liach: In the localization file, the name of spike point has a .name following, so there is an issue in the message when pairing.

CovertJaguar: Its the message that need fixing, not this.

liach: This is wrong because it is not from tiles. Will change in localization files then.

liach:
Everything else is fine right? Will fix tonight.

src/main/java/mods/railcraft/common/blocks/machine/worldspike/BlockWorldspikePoint.java
CovertJaguar: What? no that's not right either.

CovertJaguar: We are talking about the pairing messages right?

CovertJaguar: The localization files are supposed to have the ".name" suffix. getUnlocalizedName() returns the tag without the ".name" suffix. getLocalizedName() returns it with it.

Overriding getUnlocalizedName() is definitely the wrong solution for whatever you are trying to do.

CovertJaguar (approved):

src/main/java/mods/railcraft/common/blocks/machine/worldspike/TileWorldspike.java
```diff
@@ -165,13 +171,15 @@ public boolean setPoint(EntityPlayer player, WorldCoordinate coord) {
...
+                ChatPlugin.sendLocalizedChatFromServer(player, "gui.railcraft.worldspike.pair.fail.alignment", getLocalizationTag(),
+                        state.getBlock().getUnlocalizedName());
                 return false;
             }
 
             int max = getMaxPointChunks();
             if (Math.abs(xChunk - xPointChunk) >= max || Math.abs(zChunk - zPointChunk) >= max) {
-                ChatPlugin.sendLocalizedChatFromServer(player, "gui.railcraft.worldspike.pair.fail.distance", getLocalizationTag(), state.getBlock().getUnlocalizedName());
+                ChatPlugin.sendLocalizedChatFromServer(player, "gui.railcraft.worldspike.pair.fail.distance", getLocalizationTag(),
+                        state.getBlock().getUnlocalizedName());
```
CovertJaguar: Are these wrong too?

src/main/java/mods/railcraft/common/blocks/machine/worldspike/TileWorldspike.java
```diff
@@ -465,8 +492,9 @@ public NBTTagCompound writeToNBT(NBTTagCompound data) {
     public void readFromNBT(NBTTagCompound data) {
         super.readFromNBT(data);
 
-        if (needsFuel())
+        if (needsFuel()) {
```
CovertJaguar: Don't commit these style changes.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/blocks/machine/worldspike/TileWorldspike.java
liach: Just searched for usages in the block class, sorry.

src/main/java/mods/railcraft/common/blocks/machine/worldspike/TileWorldspike.java
liach: For this I probably need to remake the whole thing.

liach:
@CovertJaguar Please review again.

CovertJaguar (approved):

PR #1275, opened 2017-07-15
Let charge blocks use existing battery

### LINKED ISSUE #1224: Feeder unit stops accepting EU even when it should be

liach:
Fixes #1224
Set world early to prevent errors when NBT is read in admin feeders.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/machine/charge/TileCharge.java
+++ b/src/main/java/mods/railcraft/common/blocks/machine/charge/TileCharge.java
@@ -10,6 +10,7 @@
 package mods.railcraft.common.blocks.machine.charge;
 
 import mods.railcraft.common.blocks.charge.ChargeManager;
+import mods.railcraft.common.blocks.charge.ChargeNetwork;
 import mods.railcraft.common.blocks.charge.IChargeBlock;
 import mods.railcraft.common.blocks.machine.TileMachineBase;
 import net.minecraft.block.Block;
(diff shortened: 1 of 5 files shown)


src/main/java/mods/railcraft/common/blocks/machine/charge/TileCharge.java
```diff
@@ -10,28 +10,37 @@
...
+
     public abstract IChargeBlock.ChargeBattery getChargeBattery();
 
     private int prevComparatorOutput;
 
     public void neighborChanged(IBlockState state, World worldIn, BlockPos pos, Block blockIn) {
     }
 
+    protected IChargeBlock.ChargeBattery retrieve(Supplier<IChargeBlock.ChargeBattery> supplier) {
+        ChargeNetwork.ChargeNode node = ChargeManager.getNetwork(worldObj).chargeNodes.get(pos);
```
CovertJaguar: You should not be accessing chargeNodes directly here. Those collections should probably be private. That or we need a new function if getNode(pos) causes a stack overflow.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/blocks/machine/charge/TileCharge.java
liach: You defined the accessibility. I used it because it is public. So can you add something like `getNodeWithOutModification` that don't actively check for creating a node?

CovertJaguar (approved):

PR #1277, opened 2017-07-16
Serialize cart battery charge

liach:
Otherwise they lose power once chunk unloads.


CovertJaguar (commented):
This works, but an argument could be made that we should add write/read functions to the battery itself and just call those.

liach:
I originally wanted to add default methods, but since it is in the API (Java6), we can add static methods instead.

liach:
@CovertJaguar Fixed. You need to merge the pull in the api first.

CovertJaguar:
The API is still Java 6? I thought I gave up caring ages ago.

liach:
I think we should keep it J6...

PR #1278, opened 2017-07-16
Pretty much it for ai disabling

### LINKED ISSUE #316: Investigate Entity AI of Minecart Riding Entities for Issues
As demonstrated in the following video, apparently non-player entities riding minecarts can now move the minecart randomly while doing so. They can do this even against the pull of a Locomotive.

https://www.youtube.com/watch?v=qiFSU6qnrag

liach:
Closes #316
Handle one tick after vanilla has handled because event is in the
beginning of the update.
The pig is not able to pull for power 3, although it can resist power 1 of locos to some extent.

### DIFF
--- a/src/main/java/mods/railcraft/common/carts/MinecartRiderAIDisabler.java
+++ b/src/main/java/mods/railcraft/common/carts/MinecartRiderAIDisabler.java
@@ -10,8 +10,10 @@
 
 package mods.railcraft.common.carts;
 
+import net.minecraft.entity.Entity;
 import net.minecraft.entity.EntityLiving;
 import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.entity.item.EntityBoat;
 import net.minecraft.entity.item.EntityMinecart;
 import net.minecraftforge.event.entity.living.LivingEvent;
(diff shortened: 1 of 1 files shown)


CovertJaguar (changes requested):
Explain to me how this is more effective that my previous fix of turning off all 6 tasks? The only real change of import seems to be the % 8 to % 5.

liach:
This time, it is more like a vanilla logic (you can see how boats disable some flag in EntityLiving.class) pigs in the cart will follow my carrot, which is fine. When it goes against a creative loco, it can only make the train stop when the loco is the slowest; if the loco is faster, the pig cannot resist the loco.

CovertJaguar:
I originally based this on the boat code. 8 was chosen as its more efficient to compute and everything was disabled for completeness.

liach:
Then we don't need a cage cart :disappointed:

CovertJaguar:
The Cage cart was for flying mobs.

liach:
Closing for the ai part is actually in `onUpdate` block in EntityMinecart class. It is really a mess, and the logic is based on some sort of random trig functions.

Issue #1203, opened 2017-05-13
Get rid of CoFH's RF as a dependency/API


Generalcamo:
Edit by liach: the whole thing will be under revision since RF is alive now!
Another edit by liach: MJ is gonna be back, so kick RF away!

Redstone Flux is currently unmaintained for 1.10 (until recently) and 1.11. Current plans regarding the updates for it involve requiring mods that use the CoFH Energy API to install a separate dependency. While good for the API, this is bad for Railcraft, especially since we do not use a good portion of the API itself. This would involve requiring our users to download the outside mod, when we don't even use it that much really.

We should get rid of it, in favor of our own energy system. 
<details> <summary>Checklist</summary>

- [x] Make the Rolling Machine and Rock Crusher utilize Charge instead of RF.
- [x] Look into options for the RF Module. Move to Forge Energy?
- [ ] Separately look into options for the Engines. Should these also be moved with the RF Module, or should we make them MJ engines for Buildcraft 8 (Like they were before 1.7).
- [ ] Some sort of Basic Generator, for people without IC2 to utilize until the Charge Generators (and whatever we do with the RF module) are fully ported? Maybe a Stirling Generator that takes furnace fuel? Could also function as the basic generator for those who cannot afford a whole steam setup when everything is fully ported.
</details>

Skillotic0703:
You can make RF a soft dependency. Your machines should be able to be powered with charge when RF API isn't present. And when the RF API is loaded, your machines should be able to use both energy systems.
And also now that the RF API is installed as a separate jar file, it will be updated ASAP when future updates come!

Skillotic0703:
https://minecraft.curseforge.com/projects/redstone-flux

liach:
@Skillotic0703 That is what we are doing now. We are seeking to remove RF completely because of the imbalance brought by the RF mods.

llloyd4:
Let me see if I understand this, and if I sound stupid please forgive. >.< Does this mean the electric train powered rails and certain machines won't be using RF at all? As I was thinking of setting up some powered rails and an electric train but I couldn't even find the EU powered rail accepter thing that was in 10.1.2. Umm, is there a wiki / guide on how to use powered rails and electric trains in the 10.2.0 version of Railcraft? Or is that whole system still a WiP and I'm stuck to steam trains and/or the Minecraft vanilla carts and boosters?

liach:
use ic2 charge feeder for now.

liach:
[...]

PR #1282, opened 2017-07-16
Migrated to Forge Energy

liach:
Removed RF api interfaces, but it is kept for buildcraft API
See #1203


CovertJaguar:
I was more planning to remove entirely rather than switch to the equally horrible FU, except maybe for the RF Cart and Loaders.

AEnterprise:
sidenote: you do not have to keep any of this for the BC api, also for MC 1.11.2 there now is a maven you might want to use instead of having to update the api here manually

https://www.mod-buildcraft.com/maven/

liach:
Yeah... This is just the first step.

liach:
@CovertJaguar I urge you to merge this pull first so that we won't get crashes and complaints for our incorrectly ported CoFH RF API.

CovertJaguar:
Then just comment out RF completely. I'm not adding FU support.

PR #1284, opened 2017-07-16
Fixes #1248

### LINKED ISSUE #1248: Forestry water capsules are infinite in the steam locomotive.
Put a wax capsule of water in the input slot --> goes into the processing slot --> water goes into the loco --> capsule still containing water goes in the output slot.

liach:
Tested: the capsules are simply consumed.


liach:
This is not quite right. Closing for another fix that is combined with tank cart fixes.

PR #1285, opened 2017-07-16
Fluid and Tank cart fixes

### LINKED ISSUE #1194: Tank carts do not netsync their contents properly

### LINKED ISSUE #1248: Forestry water capsules are infinite in the steam locomotive.

### LINKED ISSUE #1270: Fluid Cart fluid issues
testing on the creative server

- [x] fluid carts won't accept fluid from items
- [x] fluid doesn't update until gui is opened

- [ ] somehow, fluid can exist/filter can be changed on a tank cart with a different filter

liach:
Fixes #1194
Fixes #1248
Fixes #1270
Now both 3 slot and 2 slot systems work
Tested on my side

### DIFF
--- a/src/main/java/mods/railcraft/common/fluids/FluidTools.java
+++ b/src/main/java/mods/railcraft/common/fluids/FluidTools.java
@@ -42,6 +42,8 @@ import javax.annotation.Nullable;
 import java.util.*;
 import java.util.function.Predicate;
 
+import static mods.railcraft.common.util.inventory.InvTools.isEmpty;
+
 /**
  * @author CovertJaguar <http://www.railcraft.info>
  */
@@ -64,6 +66,26 @@ public final class FluidTools {
         if (fluidStack == null)
             return "null";
         return fluidStack.amount + "x" + fluidStack.getFluid().getName();
+    }
+
+    /**
+     * Forge is too picking here. So no {@link InvTools#isEmpty(ItemStack)} here.
+     *
+     * @param stack The stack to check
+     * @return True if the liquid is successfully consumed
+     */
+    public static boolean isSuccessfullyConsumed(@Nullable ItemStack stack) {
+        return stack != null && stack.stackSize == 0;
+    }
+
+    /**
+     * Forge is too picking here. So no {@link InvTools#isEmpty(ItemStack)} here.
+     *
+     * @param stack The stack to check
+     * @return True if the liquid failed to drain/fill
+     */
+    public static boolean isFailed(@Nullable ItemStack stack) {
+        return stack == null;
     }
 
     @Nullable
@@ -97,7 +119,7 @@ public final class FluidTools {
 
     private static ProcessState tryFill(IInventory inv, StandardTank tank, ItemStack container) {
         ItemStack filled = FluidUtil.tryFillContainer(container, tank, Fluid.BUCKET_VOLUME, null, true);
-        if (InvTools.isEmpty(filled)) {
+        if (isFailed(filled)) {
             sendToOutput(inv);
             return ProcessState.RESET;
         }
@@ -107,12 +129,12 @@ public final class FluidTools {
 
     private static ProcessState tryDrain(IInventory inv, StandardTank tank, ItemStack container) {
         ItemStack drained = FluidUtil.tryEmptyContainer(container, tank, Fluid.BUCKET_VOLUME, null, true);
-        if (InvTools.isEmpty(drained)) {
+        if (isFailed(drained)) {
             sendToOutput(inv);
             return ProcessState.RESET;
         }
         inv.setInventorySlotContents(1, InvTools.makeSafe(drained));
-        return ProcessState.FILLING;
+        return ProcessState.DRAINING;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/mods/railcraft/common/fluids/FluidTools.java
```diff
@@ -66,6 +68,26 @@ public static String toString(@Nullable FluidStack fluidStack) {
...
+        return stack != null && stack.stackSize == 0;
+    }
+
+    /**
+     * Forge is too picking here. So no {@link InvTools#isEmpty(ItemStack)} here.
+     *
+     * @param stack The stack to check
+     * @return True if the liquid failed to drain/fill
+     */
+    public static boolean isFailed(@Nullable ItemStack stack) {
```
liach, line 87: This one fixes `isEmpty()`'s weakness that it cannot discern null stacks from used one-time container of stack size zero.

src/main/java/mods/railcraft/common/fluids/FluidTools.java
```diff
@@ -66,6 +68,26 @@ public static String toString(@Nullable FluidStack fluidStack) {
         return fluidStack.amount + "x" + fluidStack.getFluid().getName();
     }
 
+    /**
+     * Forge is too picking here. So no {@link InvTools#isEmpty(ItemStack)} here.
+     *
+     * @param stack The stack to check
+     * @return True if the liquid is successfully consumed
+     */
+    public static boolean isSuccessfullyConsumed(@Nullable ItemStack stack) {
```
liach, line 77: This is for one-time containers, although it is not really used.

src/main/java/mods/railcraft/common/fluids/FluidTools.java
```diff
@@ -107,12 +129,12 @@ private static ProcessState tryFill(IInventory inv, StandardTank tank, ItemStack
...
     private static ProcessState tryDrain(IInventory inv, StandardTank tank, ItemStack container) {
         ItemStack drained = FluidUtil.tryEmptyContainer(container, tank, Fluid.BUCKET_VOLUME, null, true);
-        if (InvTools.isEmpty(drained)) {
+        if (isFailed(drained)) {
             sendToOutput(inv);
             return ProcessState.RESET;
         }
         inv.setInventorySlotContents(1, InvTools.makeSafe(drained));
-        return ProcessState.FILLING;
+        return ProcessState.DRAINING;
```
liach, line 137: This has bugged me a lot, but I managed to fix this :smile: and one-time containers work then

src/main/java/mods/railcraft/common/fluids/FluidTools.java
```diff
@@ -198,90 +210,108 @@ public static ProcessState processContainer(IInventory inv, StandardTank tank, b
...
+    /**
+     * Process containers in input/output slot like tha in the tank cart.
+     *
+     * @param tank       Fluid tank
+     * @param inv        The inventory that contains input/output slots
+     * @param inputSlot  The input slot number
+     * @param outputSlot The output slot number
+     * @return {@code true} if changes have been done to the tank
+     */
+    public static boolean processContainers(StandardTank tank, IInventory inv, int inputSlot, int outputSlot) {
```
liach, line 223: I've tested extensively with ic2 fluid cells and forestry one-time capsules. Working as intended.

src/main/java/mods/railcraft/common/fluids/tanks/StandardTank.java
```diff
@@ -34,6 +35,8 @@ public void refresh() {
     };
     private int tankIndex;
     private boolean hidden;
+    @Nullable
+    private Consumer<StandardTank> updateCallback;
```
liach, line 39: This callback allows tank cart to get updated, and it works well.

src/main/java/mods/railcraft/common/fluids/tanks/StandardTank.java
```diff
@@ -87,9 +94,10 @@ public boolean matchesFilter(FluidStack fluidStack) {
 
     @Override
     public int fillInternal(FluidStack resource, boolean doFill) {
-        if (!matchesFilter(resource))
```
CovertJaguar: Why remove this check? Its there to prevent bugs.

src/main/java/mods/railcraft/common/fluids/tanks/StandardTank.java
```diff
@@ -98,14 +106,22 @@ public int fill(final FluidStack resource, final boolean doFill) {
             return 0;
         if (resource.amount <= 0)
             return 0;
-        return super.fill(resource, doFill);
+        if (!matchesFilter(resource))
```
CovertJaguar: This is unnecessary, its called via canFillFluidType() in super.fill().

src/main/java/mods/railcraft/common/fluids/tanks/StandardTank.java
CovertJaguar: 👍

src/main/java/mods/railcraft/common/fluids/FluidTools.java
CovertJaguar: Copypasta bug ftw!

src/main/java/mods/railcraft/common/fluids/FluidTools.java
CovertJaguar: And it will explode completely the instant you feed it a multi-bucket fluid container.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/fluids/tanks/StandardTank.java
liach: This check is not for `fillInternal` according to the superclass' javadoc. It is for `fill` only, and you put it in the wrong spot.

src/main/java/mods/railcraft/common/fluids/FluidTools.java
liach: @CovertJaguar For that claim, we need to find one of those containers and test. Extrautil doesn't have one, we need to find others!

src/main/java/mods/railcraft/common/fluids/tanks/StandardTank.java
CovertJaguar: No I didn't. I put it where it needed to be to prevent any possibility of filling with the wrong fluid.

fillInternal is for bypassing tanks with canFill = false.

liach:
@CovertJaguar Can you work on these pulls? I will not be able to devote a lot on these pulls in the following days.

CovertJaguar (approved):

liach:
@CovertJaguar I will left you decide which model you will take. I suggest you to take the 3 slot model. If you want to make the change, I suggest you changing the texture and adding the slot on your own, because I am not very good at that (Time is an issue too)

PR #1286, opened 2017-07-16
Mark runClient as never up to date

liach:
if it is marked as up to date gradle skips launching and you need to clean everything


CovertJaguar (approved):

CovertJaguar:
I don't ever use this, but sure ok.

PR #1294, opened 2017-07-29
Implemented Jukebox cart and moving whistle

### LINKED ISSUE #633: Jukebox Cart
This is simply a minecart with a jukebox, great for listening to music on the go.

liach:
Tested. It works!
Closes #633


CovertJaguar (approved):
👍 

The "now playing" message is broken and the texture is too  red, but those can be fixed later.

Issue #1207, opened 2017-05-16
[1.10.2-10.1.2] Recipes containing creosote containers don't allow buckets of creosote


MalkContent:
I glanced at the code a bit and I assume that the recipes are intended to accept them. Tried to find why the don't, but didn't find the end of that rabbithole.
Best guess: Bucket gets added only after the recipes are set up.

liach:
That is intentional. CJ opened a pull request to forge but was not accepted because he failed to maintain it.

MalkContent:
oh. strange. okay then :) i mean i tried this in a vanilla + railcraft only setting with the universal creosote bucket that was added by rc so i'm genuinely curious as to what the issue is.

liach:
The issue is with fluid containers not recognized in crafting recipe.

Generalcamo:
There are some plans to "fix" this with our own implementation of the prior pull request. I'm sure CovertJaguar can explain more though.

liach:
https://github.com/MinecraftForge/MinecraftForge/pull/3332

Generalcamo:
Affects Railcraft recipes as well. This is a bit more important than I thought.

liach:
We can create a custom recipes class for this for now.

liach:
Fixed in that commit above.

Issue #1106, opened 2017-03-09
Creosote and arthropods


liach:
Since creosote is intended to prevent bugs biting away rails, they should be able to damage arthropods. I suggest the addition of splash creosote to damage arthropods and make arthropods slowly lose health when in creosote.

Natesky9:
There are alot of improvements that could be made to the brewing system. This seems like a neat idea, but it would be a novelty item.

If there were more to this, then definitely

liach:
This will probably wait until the addition of creosote back as a fluid.

Issue #309, opened 2014-09-25
New way of making creosote wood blocks


Forecaster:
Spawned as a result of suggestion #308

Instead of a normal crafting recipe it'd be interesting if you made creosote wood blocks by throwing a log into a block of creosote fluid placed in the world.

The wood block would soak in the oil and after a time the oil will be consumed and the wood block will turn into a creosote block.

This change may be better suited for when the mixer machine is added. You could then have this as an early way of making creosote blocks (mainly for building a water tank probably) while later game using the mixer to create the blocks faster.

A machine more akin to an infuser of some sort may be more appropriate but is not warranted by a single recipe.

Forgot to add:
Credit goes to neptunepink (@purpleposeidon) for this suggestion

liach:
I can implement soaking here, but I don't know if you really want this or another machine.

bookerthegeek:
Why not do both?

PR #1321, opened 2017-08-27
Adding creosote and steam back

liach:
At the same time, I also added recipes that support fluids and cleaned up some deprecated methods.


liach:
Should I do #1106 at the same time?

liach:
How about #309? Do I add that too?
By the way, I've fixed #1207

src/main/java/mods/railcraft/common/carts/ItemLocoElectric.java
```diff
@@ -33,12 +33,10 @@ public ItemLocoElectric(IRailcraftCartContainer cart) {
     }
 
     @Override
-    public void finalizeDefinition() {
-        super.finalizeDefinition();
+    public void defineRecipes() {
+        super.defineRecipes();
         //TODO: Change to battery
```
CovertJaguar: This TODO can be removed.

src/main/java/mods/railcraft/common/util/crafting/ShapelessFluidRecipe.java
```diff
@@ -0,0 +1,168 @@
...
+     *
+     * @return The recipes input vales.
+     */
+    public List<Object> getInput() {
+        return this.input;
+    }
+
+    @Override
+    public ItemStack[] getRemainingItems(InventoryCrafting inv) {
+        return ForgeHooks.defaultRecipeGetRemainingItems(inv);
```
CovertJaguar: If this is copied directly from my PR, then this part is incomplete. It needs to find and drain the correct amount from the containers.

src/main/java/mods/railcraft/common/blocks/RailcraftBlocks.java
```diff
@@ -206,23 +191,37 @@ public void register() {
             if (!(block instanceof IRailcraftBlock))
                 throw new RuntimeException("Railcraft Blocks must implement IRailcraftBlock");
             IRailcraftBlock blockObject = (IRailcraftBlock) block;
-            blockObject.initializeDefinintion();
-            Railcraft.instance.recipeWaitList.add(blockObject);
```
CovertJaguar, line 207: Where did the recipe definitions go?

CovertJaguar (commented):
What does the Creosote potion effects do?

src/main/java/mods/railcraft/common/blocks/RailcraftBlocks.java
liach: It goes to `RailcraftModulePayload`, in one of the initializing methods.

src/main/java/mods/railcraft/common/util/crafting/ShapelessFluidRecipe.java
liach: Will adjust that. Or can I keep a list after the `getResult` call, presuming that this is always called after getting the result?

src/main/java/mods/railcraft/common/carts/ItemLocoElectric.java
liach: :+1:

liach:
As creosote is supposed to kill bugs for the wooden ties, the creosote potion effect deals damage to the arthropods with an efficiency similar to that of poison potion to players. It can kill arthropods, however. Non-arthropods will not be affected, but if an arthropod attacks you, it will get half of the length of the creosote effect on you, if there is any. When living entities stay in creosote liquid, they will get 5 seconds of creosote, which will reapply once it expires. You can put creosote bottle as an ingredient to turn mundane potion into creosote potion. It has a long version and a strong version, similar to those of the speed effect.

liach:
@CovertJaguar Please review again,,, Thanks!

src/main/java/mods/railcraft/common/util/crafting/ShapelessFluidRecipe.java
liach: Done. I checked in `matches` and used that drain results.

CovertJaguar (approved):

liach:
I will include this for 10.2.1 for it has a bugfix for fluid container recipes.

PR #1323, opened 2017-08-28
Improved bore head again

### LINKED ISSUE #1314: Can't see to enchant the bore head.
I noticed it said the 10.1.2 version now supports enchanting the bore heads.  How do you go about enchanting them?  I've tried the Anvil, the standard Table, and the Advanced Enchanting Mod but none of them seem to allow me to place an enchantment on the bore head.

liach:
Fixes #1314
Make the silk touch actually work.
Add the firing of harvest events.
Add better compatibility for tool classes.
Better enchantment with compatibilities (extending ItemTool but make them look like regular items as much as possible)


CovertJaguar (approved):
Some concern about the reflection working in all environments, but nothing that can't be fixed later if its a problem.

Issue #1204, opened 2017-05-14
Trains locking/idle not serialized?


purpleposeidon:
Locomotives have a "one of my carts is being held by a locking track, so I'll idle even though my button says I'm running" variable, right? It's probably not being serialized.

1. Locomotive at lava refill station, one of the tank carts being held. Locomotive is chill.
2. Visit nether for cart to grab more lava
3. Return, locomotive is not chill & is all sproinged out.

CovertJaguar:
"sproinged out" that's a new one.

CovertJaguar:
I think the issue is that I'm not telling the train that it is locked in this case.

CovertJaguar:
However, the code looks like it should be handling this correctly. I may need to use the debugger.

purpleposeidon:
I may also be wrong! We'll see if I can get it happening again.

liach:
@purpleposeidon Is it still happening?

purpleposeidon:
I haven't noticed it recently.

liach:
#1195?

liach:
I think it exists, since no track kit calls `TileEntity#markDirty` at all!

liach:
I've opened a pull to debug. Hope it works!

liach:
This one is actually more similar to #657.

liach:
I have confirmed that train states are indeed not serialized.

Issue #855, opened 2016-07-22
New Disposal Track features


CovertJaguar:
Boni suggested I add something similar: https://twitter.com/bonusboni/status/753678824588644352

So I got to thinking and figured, hey the Disposal Track already does this for entities, no reason we can't make it do it for items too.

So the idea:
- [x] Add a filter slot gui to limit cart types
- [x] Make Disposal Track extract items from carts
- [x] Make Disposal Track spawn extracted items or insert into inventory below track.
- [x] Rename to Dumping Track?

Forecaster:
Not sure if the "insert into inventory" part is needed. We have the unloaders for that, as well as the hopper cart being able to do that on it's own. The rest I approve of though!

CovertJaguar:
It would still be a very "dumb" system without all the bells and whistles of an unloader. No filters (besides cart), no stocking, no locking track integration/interaction, etc...

Forecaster:
Maybe if it's also slower than the normal unloader?

CovertJaguar:
Probably.

Vexatos:
I actually like the name "Disposal Track".

Forecaster:
Disposal track is great 👍

MalkContent:
How about making it dump to a side and vomit all the items in entity form at once. The cart model could tilt in that direction.
People can then build their own systems to deal with the item onslaught. Or garbage.
Also: I don't think it's necessary to have it dump mobs, too.

dj3520:
> vomit all the items in entity form at once

Massive entity count increase, also I've seen items that respond to being dropped.

MalkContent:
So use the unloaders if that's either to taxing or unwanted gameplaywise. I actually like the thought that there's items that can't be handled this way.

liach:
I've implemented all the ideas here. I just need to add a GUI and test.

liach:
This is "no longer implemented" because old code cannot be merged onto the new base any more.

CovertJaguar:
Right so I need to do some git magic it looks like to bring this in.

CovertJaguar:
Closed by 12fedb804fc8fc0acac3e9649f5a546bc6bf4697

PR #1328, opened 2017-09-04
Added dumping track features

liach:
Also tweaked locking tracks to use early world setting mechanism.
See #855 and #1204


src/main/java/mods/railcraft/common/gui/containers/RailcraftContainer.java
```diff
@@ -66,7 +66,7 @@ public void addListener(IContainerListener listener) {
     @Override
     public final void detectAndSendChanges() {
         super.detectAndSendChanges();
-        if (FMLCommonHandler.instance().getEffectiveSide() == Side.SERVER) {
+        if (FMLCommonHandler.instance().getSide() == Side.SERVER) {
```
CovertJaguar: This breaks everything. These functions have vastly different meanings. I need to know which thread I'm on, not which binary.

src/main/java/mods/railcraft/common/util/inventory/AdjacentInventoryCache.java
```diff
@@ -81,4 +81,20 @@ public InventoryComposite getAdjacentInventories() {
...
+            invs.remove(face);
+            TileEntity tile = tiles.get(face);
+            if (tile != null && (filter == null || filter.test(tile))) {
+                IInventoryObject inv = InventoryFactory.get(tile, face.getOpposite());
+                if (inv != null)
+                    invs.put(face, inv);
+            }
+
+            changedSides.clear();
+            sortedInvs.add(invs.get(face));
```
CovertJaguar: You realize you are returning an ever growing list here? Every time the function is called, it adds a new entry. You removed the clear. Its also returning _all_ the other sides too.

src/main/java/mods/railcraft/common/util/inventory/StandaloneInventory.java
```diff
@@ -59,6 +62,12 @@ public StandaloneInventory(int size, @Nullable String name, @Nullable RailcraftT
         this.callback = callback == null ? null : new TileCallback(callback);
     }
 
+    public StandaloneInventory(int size, @Nullable String name, @Nullable ITrackKitInstance callback) {
```
CovertJaguar, line 65: Why don't you just use the tile callback? You can get the tile from the track instance.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/gui/containers/RailcraftContainer.java
liach: Will revert this one.

src/main/java/mods/railcraft/common/util/inventory/AdjacentInventoryCache.java
liach: This one is actually unused. I will try to revert this change at best, since this cache can only pull inventory from all sides.

src/main/java/mods/railcraft/common/util/inventory/StandaloneInventory.java
liach: I don't want to initialize the callback until the tile of the track kit is set. That code will look ugly.

CovertJaguar: Hmm

CovertJaguar (approved):

liach:
I really want to merge this, for it fixes #1204

CovertJaguar:
You want to update this?

liach:
yeah, i will, but wait for me to be free (i have some real life stuff)

liach:
hmm, the rebase will be hard

CovertJaguar:
Easier than my last merge.

PR #1329, opened 2017-09-04
Fixes redstone signal receiving for placed signal controllers

### LINKED ISSUE #1271: Signal boxes redstone
When placed, a signal box will not see adjacent redstone until updated
This nearly led me to believe that boxes could not transmit redstone between linkages

front and the one behind are set up the same, Redstone = Red aspect
though the one in front was placed after the repeater

I do have a request on this, since the boxes have alot of potential. There is no real way to tell whether a box is powered by redstone. You can see the aspect as the green dot, but nowhere can you notice the redstone signal being passed to it, say from another box. On that as well, if all sides of a box are occupied, there is no way to tell the signal either

Can we get a signal/redstone display in the gui?

liach:
Fixes #1271
Displaying redstone state is not necessary, however.

### DIFF
--- a/src/main/java/mods/railcraft/common/plugins/forge/PowerPlugin.java
+++ b/src/main/java/mods/railcraft/common/plugins/forge/PowerPlugin.java
@@ -9,6 +9,7 @@
 package mods.railcraft.common.plugins.forge;
 
 import net.minecraft.block.Block;
+import net.minecraft.block.state.IBlockState;
 import net.minecraft.init.Blocks;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.math.BlockPos;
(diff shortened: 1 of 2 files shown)


src/main/java/mods/railcraft/common/plugins/forge/PowerPlugin.java
```diff
@@ -56,14 +56,10 @@ public static boolean isRedstonePowered(World world, BlockPos pos) {
         return false;
     }
 
-    @Deprecated
     private static boolean isRedstonePowering(World world, BlockPos pos, int yOffset, EnumFacing side) {
         BlockPos wirePos = pos.up(yOffset).offset(side);
-        Block block = WorldPlugin.getBlock(world, wirePos);
-        if (block == Blocks.REDSTONE_WIRE) {
```
CovertJaguar, line 63: Removing the wire check changes the meaning of this function.

liach:
It is for other mods, such as redpower,'s compatibility

PR #1333, opened 2017-09-11
Added bed cart

### LINKED ISSUE #779: Enhancement: Sleeper Cart
A minecart, which behaves exactly like the standard minecart (you can ride in it), only when it is night, it qualifies as a bed and it doesn't set your spawn.

liach:
Closes #779.
Sleeper means ties in en_GB, so I call the cart a bed cart.
Needs textures from CovertJaguar. You may do some change in rendering later, too.
Right now, the cart has no item texture. Its cart block is a head part of a bed.
Anything else works well.
When a player hits the sleep key, he will receive message or sleep successfully if he should.
Every cart listens to the sleep events. I hope that it would not cause too much issues with performance.

### DIFF
--- a/extras/ide/inspections/net/minecraftforge/event/entity/player/annotations.xml
+++ b/extras/ide/inspections/net/minecraftforge/event/entity/player/annotations.xml
@@ -2,4 +2,7 @@
     <item name='net.minecraftforge.event.entity.player.PlayerDestroyItemEvent net.minecraft.util.EnumHand getHand()'>
         <annotation name='org.jetbrains.annotations.Nullable'/>
     </item>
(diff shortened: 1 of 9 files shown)


liach:
@CovertJaguar What block should I render in the cart?

CovertJaguar:
This may require a custom renderer for the contents, similar to the RF cart and maintance carts.

CovertJaguar:
How exactly does this work? Does it set the spawn point?

liach:
no, it just set the sleeping player riding the cart when the player sleeps and wakes.

liach:
can you help writing a renderer?

PR #1337, opened 2017-09-18
Fixes tunnel bore destroy check

### LINKED ISSUE #1335: Tunnel Bore Destroys blocks
Just started using version 10.2.0 and seem to be having an issue with the tunnel bore destroying blocks instead of mining them.  It appears I have the correct settings in the config file.  Hoping you can look into this. 

        bore {
            # 
            # change to 'destroyBlocks=true' to cause the Bore to destroy the blocks it mines instead of dropping them
            # 
            B:destroyBlocks=false

            # 
            # change to 'mineAllBlocks=false' to enable mining checks, use true setting with caution, especially on servers
            # 
            B:mineAllBlocks=true

            # 
            # add block ids to 'mineableBlocks' in a common separated list to define non-vanilla blocks mineable by the tunnel bore 
            # ignored if 'tweaks.carts.bore.mineAllBlocks=true' 
            # metadata sensitive entries can be defined in the form 'modid:blockname#metadata' 
            # Example:mineableBlocks= { minecraft:stone, minecraft:stonebrick#3 }
            # 
            S:mineableBlocks={}

            # 
            # adjust the speed at which the Bore mines blocks, min=0.1, default=1.0, max=50.0
            # 
            D:miningSpeed=1.0
        }

liach:
fixes #1335


CovertJaguar:
@liach Why is this set to merge into your branch? Looks good btw, just not sure what to do with it.

CovertJaguar (approved):

liach:
Because I thought the main branch was no longer active.

liach:
Will put it in main branch then, I assume.

liach:
Ah yes, I remember why the base was my branch. Sorry, but I can only merge it into my dev branch for now.

PR #1430, opened 2018-06-03
[1.12.2] Fix Shapeless Craft in Rolling Machine

xfl03:
Both Manual Rolling Machine & Rolling Machine has this problem.

**Before** fix:

**After** fix:


liach (approved):
真心了，居然这个炸了

xfl03:
Thanks!

PR #1438, opened 2018-06-12
[1.12.2] Fix some GUI Text

xfl03:
Involved GUI:
- Trade Station
- Cart Undercutter
- ALL Detectors

Before Fix:

After Fix:


liach:
good good good. next time say involved instead of influenced

liach (approved):

PR #1484, opened 2018-09-08
Implement steam boiler models

adrwa:
And also fix them not forming a multiblock


liach (approved):
Thank you for spending hours helping!

PR #1486, opened 2018-09-08
Fix tanks not forming a multiblock

adrwa:


adrwa:
Just looked back and I'm not sure why this even fixed it, it changes nothing.. :/

PR #1500, opened 2018-09-20
Fixed Ic2 classic compat.

Trinsdar:
Lang files don't work however, because in the lang files mfsu is lowercase but in the unlocalized name mfsu is uppercase. Either the lang files need to be changed or the unlocalized name needs to be changed. It would probably be easier to change the unlocalized name.


liach:
Ehh wrong branch, indev 1.12.2 instead

liach (approved):

Trinsdar:
do you want me to change the base branch to indev-1.12?

liach:
@Trinsdar yes.

PR #1513, opened 2018-10-17
Indev 1.12.2

CovertJaguar:
Testing the travis app mostly.


liach:
@CovertJaguar The main issue in the two recipe classes is that we cannot make sure if recipe users check matches before getting remaining items.

src/main/java/mods/railcraft/common/util/crafting/RemainingItemShapelessRecipe.java
```diff
@@ -73,7 +74,7 @@ public ItemStack getRecipeOutput() {
 
     @Override
     public NonNullList<ItemStack> getRemainingItems(InventoryCrafting inv) {
-        if (lastResult == null) {
+        if (lastInv != inv) {
             matches(inv, null);
```
liach, line 78: See https://pastebin.com/iK3V4RLX
https://github.com/raoulvdberge/refinedstorage/blob/ec89e43e35fd039d9aa2ae04182a2caaa97733e3/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java#L227

I'd assume modders aren't malicious enough to edit the grid between matches and getRemainingItems calls.

liach (approved):

CovertJaguar:
hmm... I have no idea what this checks tab is supposed to look like...

liach:
I think it is for travis ci and other things like cla

liach:
Consider squash merging this.

liach:
ah, travis ci indeed broke somehow

liach:
Will get an alternative fix on the recipe.

CovertJaguar:
More Travis testing

liach:
@CovertJaguar Appears you need to remove the old continuous integration travis ci

CovertJaguar:
Looks like it works!

liach:
Hmm, should I move the fix to rebornstorage to another branch? This one has too many commits and is not always right if people modify the grid and get remainders without doing matches check. Should I make sure things are totally right at the cost of performacne?

CovertJaguar:
I haven't even looked at the code or problem, it was just a convenient branch for testing Travis. But yes, your private branches should probably be deprecated in favor of more targeted temporary branches.

PR #1516, opened 2018-10-21
Simplify Steam Turbine and fix Flux Transformers

Generalcamo:
This simplifies the Steam Turbine to utilize Charge Units internally, rather than MJ internally. It also adds support for Forge Energy for generation compatibility with other mods implementing that standard.

It also fixes a few issues with Flux Transformers and the FE Plugin when those were ported from RF.


src/main/java/mods/railcraft/common/plugins/forge/EnergyPlugin.java
```diff
@@ -78,7 +78,7 @@ public static int pushToTiles(TileEntity te, AdjacentTileCache tileCache, int pu
             if (source != null && canTileReceivePower(tile, side.getOpposite())) {
                 IEnergyStorage receiver = tile.getCapability(ENERGY, side.getOpposite());
                 if (receiver != null) {
-                    int amountToPush = source.extractEnergy(pushPerSide, false);
+                    int amountToPush = source.extractEnergy(pushPerSide, true);
                     if (amountToPush > 0) {
                         int amountPushed = receiver.receiveEnergy(amountToPush, false);
```
liach, line 83: Does this one need change, too?

Generalcamo: No, the receiving shouldn't be simulated.

liach: Ah, true means simulating. This is indeed inconvenient

liach (approved):

src/main/java/mods/railcraft/common/plugins/forge/EnergyPlugin.java
CovertJaguar: Yeah, its generally bad practice to use booleans for this, the general recommendation is a enum. But we didn't write this so we have to deal with it.

Issue #1529, opened 2018-11-04
Coupling Track crash NullPointerException while passing Carts


alwaeles:
Hello,

I tried to link carts using a Coupling Track but when I passed the carts it crashed.

I'm using Railcraft on development environment (GitHub, mc-1.12.2).

Here's the crash report: [crash-2018-11-04_13.43.15-server.txt](https://github.com/Railcraft/Railcraft/files/2546051/crash-2018-11-04_13.43.15-server.txt)

Thank you,
Axelandre

Issue #1528, opened 2018-11-03
Formed Steam Oven gives ArrayIndexOutOfBoundsException when breaking blocks.


alwaeles:
Hi back,

Still trying things. Now I built and opened a Steam Oven but when I broke one of its blocks, the game crashed.

I'm still using the same version: runClient in mc-1.12.2.

Here's the crash report: [crash-2018-11-03_16.21.34-server.txt](https://github.com/Railcraft/Railcraft/files/2545085/crash-2018-11-03_16.21.34-server.txt)

Thank you,
Axelandre

PR #1530, opened 2018-11-04
Fixed #1527 to #1529

### LINKED ISSUE #1527: Water Tank gives StackOverflowError after right clicked on.
Hi,

I tried to build a Water Tank but at the first try nothing happened by right clicking it.
Then I broke some blocks of the structure, I placed them again and I right clicked and I got a crash that occurs every time I try to play on the world.

I'm using Railcraft on development environment (GitHub, mc-1.12.2).

Here's the crash report: [crash-2018-11-03_11.12.34-server.txt](https://github.com/Railcraft/Railcraft/files/2544943/crash-2018-11-03_11.12.34-server.txt)

Thank you.
Axelandre

liach:
#1527 wrong reference to standalone inventory
#1528 edge cases for isMaster not updated when master block is set to null
#1529 null check for tagged cart on coupler

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/multi/IMultiBlockTile.java
+++ b/src/main/java/mods/railcraft/common/blocks/multi/IMultiBlockTile.java
@@ -22,6 +22,11 @@ public interface IMultiBlockTile extends IOwnable, ITile {
 
     boolean isStructureValid();
 
+    /**
+     * Note: Must call {@link #isStructureValid()} before doing master functions for edge cases!
+     *
+     * @return True if this block entity is the master block
+     */
     boolean isMaster();
 
     @Nullable
--- a/src/main/java/mods/railcraft/common/blocks/multi/TileSteamOven.java
+++ b/src/main/java/mods/railcraft/common/blocks/multi/TileSteamOven.java
@@ -159,6 +159,9 @@ public final class TileSteamOven extends TileMultiBlockOven implements ISidedInv
     public void update() {
         super.update();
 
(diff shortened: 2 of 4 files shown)


src/main/java/mods/railcraft/common/blocks/multi/TileMultiBlock.java
```diff
@@ -454,6 +454,7 @@ public final boolean isStructureValid() {
     @Override
     public final @Nullable TileMultiBlock getMasterBlock() {
         if (masterBlock != null && !isStructureValid()) {
+            masterBlock.isMaster = false;
```
CovertJaguar: Don't like this. Its not the slaves job to determine whether a master is a master.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/blocks/multi/TileMultiBlock.java
liach: So should I call `isStructureValid()` in `isMaster()` check?

liach: Updated. Added a contract to call both methods if work should be done.

CovertJaguar (approved):

CovertJaguar:
This seems safer, but I'm still thinking I need to completely redesign the logic for multiblocks so this stuff needs less boilerplate.

liach:
thanks!

PR #1539, opened 2018-11-08
Make sure all master functions are performed only after the structure…

liach:
… is confirmed to be valid
This is to resolve the followup issue where formed coke ovens access invalid inventories on game launch.


CovertJaguar:
I fixed it before I saw this.

PR #1544, opened 2018-11-24
Somehow fix compile for guys who want to build directly

liach:
**The Issue**
 - Build fails because of a line of redundant import.
 
**The Proposal**
 - Remove that useless import.

**Possible Side Effects**
 - None.
 
**Alternatives**
 - Let CJ fix it in code; (has been a few commits without realizing this issue)


liach:
Hmm, wtf did github do, creating 2 branches?

PR #1569, opened 2018-12-17
Add vm args (so players can change memory) and enable assertions for dev

liach:

**The Issue**
 - Assertions are not enabled by default.
 - Dev-runners complain that they cannot change the RAM
 
**The Proposal**
 - Add run args in gradle to fix this issue
 
**Possible Side Effects**
 - The args are from https://redd.it/5jhuk9, should be valid
 
**Alternatives**
 - No change.


build.gradle
```diff
@@ -54,6 +54,10 @@ println 'Environment: ' + project.environment
 println '*************************************************************'
 println ''
 
+// Change the 4G in -Xmx4G to your desired max RAM and 4G in -Xms4G to your desired initial RAM
+def optimizedVmArgs = [ '-XX:+UseG1GC', '-Xmx4G', '-Xms4G', '-XX:+UnlockExperimentalVMOptions', '-XX:G1NewSizePercent=20', '-XX:G1ReservePercent=20', '-XX:MaxGCPauseMillis=50', '-XX:G1HeapRegionSize=32M' ]
+def combinedVmArgs = [['-ea'], optimizedVmArgs].flatten()
```
Chocohead: You might want to add `-da:io.netty...` too if you're enabling assertions. Netty is a little careless with its assertions (or Minecraft/Forge isn't using it quite right) so mods using `IGuiHandler` can end up with assertion errors from within Netty.

CovertJaguar: You _do not_ want enable assertions for anything but our code. Everything breaks.

CovertJaguar: This is my recommendation, but I'm not sure about the syntax here.
```
def combinedVmArgs = [['-ea:mods.railcraft...', '-ea:net.minecraft...', '-ea:net.minecraftforge...'], optimizedVmArgs].flatten()
```

CovertJaguar (changes requested):

CovertJaguar (approved):

build.gradle
Chocohead: Neither Minecraft nor Forge use assertions so you're not gaining anything by adding them to the list, but neither are you losing I suppose

PR #1570, opened 2018-12-18
Fix tank valve states and multiblock rendering when valid

liach:
**The Issue**
 - Multiblocks are considered valid on the server (tested via mag glass) but don't show up as valid and do not show up gui.
 - The tank valves have 2^6, or 64, states while it just needs 4.
 
**The Proposal**
 - Changed tank valves to use only 4 states
 - Added assertions and make sure the states are updated properly with multiblocks.
 
**Possible Side Effects**
 - Might hurt your ongoing progress on moving the multiblock patterns into logics.
 
**Alternatives**
 - Leaving the tank valves with 64 states is too costly, so I changed it
 - Prompted a todo message so that if some block need off hand functionality we can add it.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/multi/MultiBlockPattern.java
+++ b/src/main/java/mods/railcraft/common/blocks/multi/MultiBlockPattern.java
@@ -245,6 +245,7 @@ public final class MultiBlockPattern {
     public enum State {
 
         VALID(TileMultiBlock.MultiBlockState.VALID, "railcraft.multiblock.state.valid"),
+        // TODO map to untested?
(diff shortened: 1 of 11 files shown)


src/main/java/mods/railcraft/common/blocks/multi/MultiBlockPattern.java
```diff
@@ -245,6 +245,7 @@ public String toString() {
     public enum State {
 
         VALID(TileMultiBlock.MultiBlockState.VALID, "railcraft.multiblock.state.valid"),
+        // TODO map to untested?
```
CovertJaguar, line 248: Possibly, it would result in spamming the retest until it succeeds though.

CovertJaguar (approved):
Looks good. It shouldn't have too much impact on StructureLogic. I already made some of the same changes.

Issue #1571, opened 2018-12-18
SIlk touch drops


bookerthegeek:
# Heyo # 
```
[05:07:29] [Client thread/ERROR] [railcraft]: Cannot initialize silk touch drops
[05:07:29] [Client thread/ERROR] [railcraft]: java.lang.NoSuchMethodException: net.minecraft.block.Block.getSilkTouchDrop(net.minecraft.block.state.IBlockState)
[05:07:29] [Client thread/ERROR] [railcraft]: java.lang.Class.getDeclaredMethod(Unknown Source)
[05:07:29] [Client thread/ERROR] [railcraft]: mods.railcraft.common.plugins.forge.HarvestPlugin.<clinit>(HarvestPlugin.java:53)
[05:07:29] [Client thread/ERROR] [railcraft]: mods.railcraft.common.items.ItemSteelAxe.initializeDefinition(ItemSteelAxe.java:41)
```
## **To Reproduce** ##
1. Install forge-1.12.2-14.23.5.2796-universal
1. Install railcraft-12.0.0-alpha-9
2. Look in the logs

### **Environment** ###
 - railcraft-12.0.0-alpha-9
 - forge-1.12.2-14.23.5.2796-universal

liach:
This thing is intended for tunnel bores to mine with silk touch enchantment. Should not impact your daily lives by too much in theory.

CovertJaguar:
What is this stuff? I see references to #1314 #1323

https://github.com/Railcraft/Railcraft/blob/fbb2d6d835aaa063e1129907f6b261b10763f948/src/main/java/mods/railcraft/common/plugins/forge/HarvestPlugin.java#L49-L62

Couldn't you do this with an AT?

liach:
Hmm, is `Game.DEVELOPMENT_ENVIRONMENT` initialized incorrectly? This cannot be done with access transformers because this method is protected (overriding methods cannot be all transformed to be public).

CovertJaguar:
I think it looks at the name of Entity.world. I don't know if its current.

CovertJaguar:
You should use [ReflectionHelper](https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/src/main/java/net/minecraftforge/fml/relauncher/ReflectionHelper.java). It does most of this for you.

Also 
[...]

PR #1573, opened 2018-12-18
Update HarvestPlugin.java

liach:
**The Issue**
 - #1571
 
**The Proposal**
 - Change to use the correct field
 
**Possible Side Effects**
 - Need to go through all other `Game.DEVELOPMENT_ENVIRONMENT` usage to see if those are correctly used. 

**Alternatives**
 - Leave it as it is; bug would be still there


CovertJaguar:
It might require a Forge update, but this should be rewritten to use ObfuscationReflectionHelper.

And all our other ReflectionHelper uses should be remapped to ObfuscationReflectionHelper.

liach:
We should not try to look for both names at once. In my opinion, we should just look for the right one. However, 

> Need to go through all other Game.DEVELOPMENT_ENVIRONMENT usage to see if those are correctly used.

This is still on our todo list.

PR #1585, opened 2018-12-25
Allow getting the texture to draw over the analog widget needle from elsewhere.

Vexatos:
Addons that want to use the fancy analog widget for their own machines might have the widget elsewhere on the GUI texture, meaning the texture to draw over the needle might be elsewhere. This allows addon devs to use the analog widget without having to copy the class only to change a single line.


CovertJaguar:
Unless you plan to rescae the needle as well, you should probably remove the ou and ov variables.

Vexatos:
That is because on my texture, the position of the widget's background texture is not identical to its position in the UI due to it not always being added, that's why both ox/oy and ou/ov are needed.

CovertJaguar (approved):

Issue #1582, opened 2018-12-25
1.12.2 missing texture


Nagyur:
**Missing texture **
Iron and steel valves has no textures.

**To Reproduce**
1. install mod
2. craft a valve

**Expected behavior**
Has some decent texture on valves. 

**Screenshots & Video**
![railcraft](https://user-images.githubusercontent.com/46134085/50424056-d67d6180-085e-11e9-83a5-d997f9f6ca08.png)

**Environment**
 - Version [e.g. Railcraft 12.0.0-beta-1]
 - forge-1.12.2-14.23.5.2796
- a bunch of other mods

liach:
Seems the tinted cube model has some issues.

Nagyur:
I checked inside the .jar the base texture is there. but i couldn't do anything more than this to fix it. Because Railcraft not referencing textured directly but shade a base one. (as far as i understand)

liach:
Yeah, the texture is fine but the model that binds the texture is not.

CovertJaguar:
I managed to get it to stop tossing errors, but it still fails to load the textures for some reason.

liach:
i got the cause and may push soon.

liach:
This is my bad.

liach:
Done with rendering, though the translation is still a todo.

PR #1590, opened 2018-12-25
Some work

liach:
Fix valve texture, Game.DEVELOPMENT_ENVIRONMENT -> DEVELOPMENT_VERSION, trying to add a log4j config but did not actually add (commented out)

rename reason: i always think that means dev env or notch name, which is what obfuscated field indicates.

Unfortunately i don't have much time to work, so this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -149,16 +149,45 @@ processResources {
     }
 }
 
+//tasks.genIntellijRuns.doLast {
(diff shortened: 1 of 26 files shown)


liach:
For the valve, see #1582

liach:
btw i also optimized blast furnace.

CovertJaguar (approved):

PR #1612, opened 2019-01-02
Added not ic2 classic check to slag dust macerator recipe

Trinsdar:
**The Issue**
If Ic2 Classic is loaded a recipe for slag dust is added that uses the no use item from ic2 classic.
 
**The Proposal**
I added a check for ic2 classic to that recipe so it doesn't get added if ic2 classic is loaded.
Granted, the recipe was not accessible, but it still was annoying me.
 
**Possible Side Effects**
none that I'm aware of.
 
**Alternatives**
I did consider just removing the recipe in my mod but I couldn't seem to get it to work and this seemed not only easier but would also not require my mod to be loaded.

Also by the way I added the complete .idea folder to the .gitignore file as it was only ignoring certain files which meant files in my local repo wanted to be added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ README.txt
 .idea/modules.xml
 .idea/gradle.xml
 .idea/compiler.xml
+.idea
 *.iml
 run/
 private/
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@ README.txt
 .idea/modules.xml
 .idea/gradle.xml
 .idea/compiler.xml
-.idea
 *.iml
 run/
 private/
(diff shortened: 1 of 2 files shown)


liach (approved):

CovertJaguar:
The .idea folder is intentionally included as part of the repo.

Thanks for fixing that @liach

CovertJaguar:
@Trinsdar it might make sense to modify ModItems to not return the no use item.

Trinsdar:
I don't think it's that simple as whenever a mod references an ic2 item via the api ic2 classic will redirect that reference to the nouse item if it is not an item in ic2 classic. This way was easier and simpler.

CovertJaguar:
Yes, ModItems is a wrapper around that. It should be able to catch that state and return an empty ItemStack instead. Which would then trigger all of Railcraft's standard recipe validation routines and automatically discard the recipes.

I'd do it, but I don't know what the resource id for the no use item is.

https://github.com/Railcraft/Railcraft/blob/f2a41baa1b96d03867e1d0c269f0df79d9c48211/src/main/java/mods/railcraft/common/items/ModItems.java#L109-L121

Trinsdar:
if you mean the item id it's ic2:itemnouse. if you mean the itemstack reference well that's in the core ic2 folder which would mean you would need ic2 classic in the libs folder, which isn't exactly very good for compat with ic2 experimental. the reference in the core ic2 folder is Ic2Items.noUse. not that that Ic2Items is different from the api IC2Items as it's got a lower case letter

liach:
ic2:itemnouse is good enough for us. Thanks!

Trinsdar:
no problem

PR #1631, opened 2019-01-09
Some resource for debugging and running in dev env

liach:
**The Issue**
 - Forge gradle's intellij idea run task is usually glitchy. It frequently fails to copy the args and does not set up the module to run properly.
 
**The Proposal**
 - Generate more idea run tasks beyond the ones provided by fg
 - The new tasks include:
  - VM args suggested by cpw
  - assertion vm args
  - log4j config to enable debug logging in vanilla
  - set the module to `railcraft_main` so that you don't need to set again
  - if the `password.properties` exist, the generated run configuration for client will automatically log you in to mojang services
  - automatically generate eula file for server launch and specify `nogui` (not showing the swing gui for dedicated server)
 - also fixed a few poor code usages in the project found during browsing
 
**Possible Side Effects**
 - The original forgegradle run tasks are still preserved as `Minecraft Client/Server`, while the new ones are called the same as the gradle task name `runClient`/`runServer`. If people dislike new ones, they can still use the old tasks.
 
**Alternatives**
 - Tweaking the existing task may cause side effects, so I chose to add new tasks.

### DIFF
--- a/src/main/java/mods/railcraft/common/util/collections/ItemKey.java
+++ b/src/main/java/mods/railcraft/common/util/collections/ItemKey.java
@@ -12,22 +12,23 @@ package mods.railcraft.common.util.collections;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraftforge.oredict.OreDictionary;
+import net.minecraftforge.registries.IRegistryDelegate;
 
 /**
  * @author CovertJaguar <http://www.railcraft.info>
  */
 public final class ItemKey {
 
-    public final Item item;
+    public final IRegistryDelegate<Item> item;
     public final int metadata;
 
     public ItemKey(Item item) {
-        this.item = item;
+        this.item = item.delegate;
         this.metadata = -1;
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/mods/railcraft/common/util/collections/ItemKey.java
```diff
@@ -12,22 +12,23 @@
...
 import net.minecraftforge.oredict.OreDictionary;
+import net.minecraftforge.registries.IRegistryDelegate;
 
 /**
  * @author CovertJaguar <http://www.railcraft.info>
  */
 public final class ItemKey {
 
-    public final Item item;
+    public final IRegistryDelegate<Item> item;
```
CovertJaguar, line 22: What is this and why is it necessary?

liach: This is to handle cases of registry substitution e.g. when a mod replaces a vanilla item. Forge has recommended doing so because mods may replace vanilla objects just like how we replace vanilla carts. This is also done in fluid container handling in the `fluid` package as well.

Issue #1601, opened 2018-12-28
Crash on startup after disabling structures module


laserlemons:
**Description of the Bug**
The game crashes on startup after disabling the structures module.
https://pastebin.com/YBCR4zNb

**Environment**
 - Railcraft 12.0.0-beta-2
 - Forge version 2796
 - I also have the whole suite of COFH mods. (Thermal Expansion)

liach:
This has been confirmed by a few other users as well.
https://github.com/Railcraft/Railcraft/blob/5af3e73e576e099d8676e621fae3d7495ab1d2c0/src/main/java/mods/railcraft/common/blocks/detector/BlockDetector.java#L170
Should be replaced by an `isLoaded()` fix.

Issue #1549, opened 2018-11-29
Cannot pipe water into a Fluid Loader or fill into cart


instinxx:
**Description of the Bug**
Unable to pipe water into a fluid loader from any side. I can insert water with a bucket by hand.

cart doesn't load from fluid loader.

**To Reproduce**
attach pipe from water source (In my case 2 different infinite water source blocks.)

**Expected behavior**
Water would flow into loader and quench the thirsty steam engine.

**Screenshots & Video**
Tried using pipes from Cyclic, immersive engineering, Ender IO and Mekenism
![image](https://user-images.githubusercontent.com/1753699/49194495-6956d900-f351-11e8-8fec-9025b01e9bea.png)

**Environment**
Railcraft 12.0.0-alpha-7
 1.12.2
 Unknown at this time. 359 possible other mods.

**Additional context**

anar4732:
I confirm this bug.
i tried immersive engineering.
RC: Alpha 7 Build 24.11.2018

liach:
After long investigation, I am finally able to fix this.

Jarrose37:
> After long investigation, I am finally able to fix this.

What did you do to fix it?

liach:
though the loaders are intended to be unidirectional, it accidemtly faces the opposite side of player facing when placed (like dispenser/piston facing)

Jarrose37:
> though the loaders are intended to be unidirectional, it accidemtly faces the opposite side of player facing when placed (like dispenser/piston facing)

IDK if I should make a new thread but I cant pump water or use water buckets to fill the fluid loader. I've tried mekanism, TE, buildcraft pipes no water goes in from any angle and when I put a water bucket in it just pops out and does not fill the fluid loader. any suggestions?

liach:
set a filter in it.

ExDomino:
Still not working!

liach:
@ExDomino how

PR #1641, opened 2019-01-11
Fixes two high-priority bugs 1601 and 1549

liach:
**The Issue**
 - #1601 #1549
 - A few other associated bugs identified (see message of the last commit)
 
**The Proposal**
 - Refactors to fix logical loopholes.

**Possible Side Effects**
 - The fluid manipulators have been tested and should be fine.
 - Some code review should be helpful.
 
**Alternatives**
 - None.


src/main/java/mods/railcraft/common/blocks/machine/manipulator/TileFluidLoader.java
```diff
@@ -54,7 +52,7 @@ public ManipulatorVariant getMachineType() {
     }
 
     @Override
-    public EnumFacing getFacing() {
+    public EnumFacing getDefaultFacing() {
```
CovertJaguar, line 55: Probably shouldn't change this, fluid loaders can't rotate.

At least point getFacing() at getDefaultFacing().

src/main/java/mods/railcraft/common/blocks/machine/manipulator/TileFluidManipulator.java
```diff
@@ -98,6 +99,7 @@ public boolean blockActivated(EntityPlayer player, EnumHand hand, EnumFacing sid
 
     @Override
     public boolean canHandleCart(EntityMinecart cart) {
+        // TODO opposite facing?
```
CovertJaguar: Probably.

src/main/java/mods/railcraft/common/blocks/machine/manipulator/TileFluidManipulator.java
```diff
@@ -113,11 +115,20 @@ protected void upkeep() {
...
     public boolean isItemValidForSlot(int slot, ItemStack stack) {
         if (slot == SLOT_INPUT) {
-            FluidStack filter;
-            return FluidItemHelper.isContainer(stack) && (FluidItemHelper.isEmptyContainer(stack) || (filter = getFilterFluid()) == null || FluidItemHelper.containsFluid(stack, filter));
+            boolean b1 = FluidItemHelper.isContainer(stack);
+            boolean b2 = FluidItemHelper.isEmptyContainer(stack);
+            FluidStack filter = getFilterFluid();
+            boolean b3 = filter == null;
+            boolean b4 = FluidItemHelper.containsFluid(stack, filter);
+            return b1 && (b2 || b3 || b4);
```
CovertJaguar: This is less efficient because it looses the short circuit property of Boolean logic.

src/main/java/mods/railcraft/common/fluids/tanks/StandardTank.java
```diff
@@ -105,8 +106,11 @@ public boolean canFillFluidType(FluidStack fluid) {
     }
 
     public boolean matchesFilter(@Nullable FluidStack fluidStack) {
+        if (fluidStack == null) {
```
CovertJaguar: We actually need to be able to set the tank contents to null.

CovertJaguar (approved):

src/main/java/mods/railcraft/common/fluids/tanks/StandardTank.java
CovertJaguar: Fixed in merge.

src/main/java/mods/railcraft/common/blocks/machine/manipulator/TileFluidManipulator.java
CovertJaguar: Much easier to read though.

src/main/java/mods/railcraft/common/blocks/machine/manipulator/TileFluidLoader.java
liach: many places access the facing field instead of calling this method. as a result, it's safer to initialize the facing field to default facing and limit rotations instead.

src/main/java/mods/railcraft/common/blocks/machine/manipulator/TileFluidManipulator.java
liach: I did that change during testing. Will revert.

src/main/java/mods/railcraft/common/blocks/machine/manipulator/TileFluidLoader.java
CovertJaguar: Had to fix this to make it compile against my changes to ITileRotate.

Issue #1491, opened 2018-09-10
Track Relayer should replace underlying track and ignore applied track kits.


AlexJ6301:
Scenario: a rail system that has iron track with track kits applied in various places (e.g. locking tracks) is to be upgraded to electric track.  At the moment the iron tracks with the kits are only recognised by the track relayer if a full rail with the track kit applied is used as the template (i.e. kind of how it used to work).

It would be great if it just replaced the underlying track, leaving the track kits, **and their underlying state**.  This is important because in a signalled system, the locking tracks etc will be changing state as the train goes over them (think about train lockdown mode, for instance - the train might not be fully past when the track relayer cart goes over the locking track, but it needs to stay "unlocked" for the whole train, other wise it will lock in place, probably across two signals blocks...)

Generalcamo:
Related to #1197

CovertJaguar:
Note to self, check status of this.

PR #1646, opened 2019-01-13
Fixes hopper cart item collection

liach:
**The Issue**
 - #1491
 
**The Proposal**
 - Added back hopper cart delay and disabling (now works properly)
 - Formats code.
 - Fixes metals chest cart not able to transfer items (it lacks an interface)
 - Some in-code utilities added.
 
**Possible Side Effects**
 - None observed.
 
**Alternatives**
 - Leaving it as it is?

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/logic/InventoryLogic.java
+++ b/src/main/java/mods/railcraft/common/blocks/logic/InventoryLogic.java
@@ -41,7 +41,7 @@ public abstract class InventoryLogic extends Logic implements IInventoryImplemen
 
     @Override
     public boolean isUsableByPlayer(EntityPlayer player) {
-        return adapter.isUsableByPlayer(player);
+        return true; //adapter.isUsableByPlayer(player); will cause infinite loop
     }
 
     @Override
--- a/src/main/java/mods/railcraft/common/blocks/logic/InventoryLogic.java
+++ b/src/main/java/mods/railcraft/common/blocks/logic/InventoryLogic.java
@@ -41,7 +41,7 @@ public abstract class InventoryLogic extends Logic implements IInventoryImplemen
 
     @Override
     public boolean isUsableByPlayer(EntityPlayer player) {
-        return true; //adapter.isUsableByPlayer(player); will cause infinite loop
+        return adapter.isUsableByPlayer(player);
     }
 
     @Override
--- a/src/main/java/mods/railcraft/common/carts/CartBaseContainer.java
+++ b/src/main/java/mods/railcraft/common/carts/CartBaseContainer.java
@@ -227,7 +227,7 @@ public abstract class CartBaseContainer extends EntityMinecartContainer implemen
 
     @Override
     public boolean isItemValidForSlot(int index, ItemStack stack) {
-        return invMappers.stream().filter(m -> m.containsSlot(index)).anyMatch(m -> m.filter().test(stack));
+        return invMappers.stream().filter(m -> m.containsSlot(index)).allMatch(m -> m.filter().test(stack));
     }
 
     /**
--- a/src/main/java/mods/railcraft/common/carts/CartBaseLogicChest.java
+++ b/src/main/java/mods/railcraft/common/carts/CartBaseLogicChest.java
@@ -11,7 +11,6 @@
 package mods.railcraft.common.carts;
 
 import mods.railcraft.common.gui.EnumGui;
-import mods.railcraft.common.util.inventory.IInventoryImplementor;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.entity.item.EntityMinecart;
 import net.minecraft.entity.player.EntityPlayer;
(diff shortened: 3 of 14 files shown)


src/main/java/mods/railcraft/common/carts/CartBaseContainer.java
```diff
@@ -227,7 +227,7 @@ public Container createContainer(InventoryPlayer playerInventory, EntityPlayer p
 
     @Override
     public boolean isItemValidForSlot(int index, ItemStack stack) {
-        return invMappers.stream().filter(m -> m.containsSlot(index)).anyMatch(m -> m.filter().test(stack));
+        return invMappers.stream().filter(m -> m.containsSlot(index)).allMatch(m -> m.filter().test(stack));
```
liach, line 230: This handles the case when there is no inventory mapper; it will return true straight.

src/main/java/mods/railcraft/common/carts/CartBaseLogicChest.java
```diff
@@ -22,7 +23,7 @@
 import net.minecraft.world.IInteractionObject;
 import net.minecraft.world.World;
 
-public abstract class CartBaseLogicChest extends CartBaseLogic implements IInteractionObject {
+public abstract class CartBaseLogicChest extends CartBaseLogic implements IInteractionObject, IInventoryImplementor {
```
CovertJaguar: I think what its actually missing is the capability function implementations.

src/main/java/mods/railcraft/common/carts/EntityCartHopper.java
```diff
@@ -140,38 +124,75 @@ public double getZPos() {
...
+                pushCounter = 0;
+            lastPosition = nowPos;
+        }
+    }
+
+    private void tryPushItem() {
+        boolean emptySlot = false;
+        // Push full stacks whenever possible
+        for (IExtInvSlot slot : InventoryIterator.get(this)) {
+            if (slot.getStack().getCount() == slot.maxStackSize()) {
```
CovertJaguar: You might want to consider the ItemStack max stack size well, not just the slot max capacity.

src/main/java/mods/railcraft/common/carts/CartBaseLogicChest.java
CovertJaguar: I think what its actually missing is the IItemHandler capability function implementations.

CovertJaguar (approved):

src/main/java/mods/railcraft/common/carts/EntityCartHopper.java
liach: changed locally, will push

src/main/java/mods/railcraft/common/blocks/logic/InventoryLogic.java
```diff
@@ -41,7 +41,7 @@ protected void dropItem(ItemStack stack) {
 
     @Override
     public boolean isUsableByPlayer(EntityPlayer player) {
-        return adapter.isUsableByPlayer(player);
+        return true; //adapter.isUsableByPlayer(player); will cause infinite loop
```
CovertJaguar: How? This is kind of necessary or you can open GUIs from across the map.

src/main/java/mods/railcraft/common/util/inventory/IInventoryImplementor.java
```diff
@@ -23,6 +23,9 @@
 public interface IInventoryImplementor extends IInventory, IInventoryComposite {
     IInventory getInventory();
 
+    @Override
+    boolean isUsableByPlayer(EntityPlayer player);
```
CovertJaguar, line 42: Redundant?

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/blocks/logic/InventoryLogic.java
liach: But apparently the standalone inv delegates this method call back to its owner, so the current implementation would cause a stack overflow error.

src/main/java/mods/railcraft/common/util/inventory/IInventoryImplementor.java
liach: This serves as a reminder that implementors must implement this method instead of delegating.

CovertJaguar (approved):

src/main/java/mods/railcraft/common/blocks/logic/InventoryLogic.java
CovertJaguar: But the adapter points at a helper function in TileRailcraft and its own implementation for carts, neither of which refer to the standalone inv.

src/main/java/mods/railcraft/common/util/inventory/IInventoryImplementor.java
CovertJaguar: Fair enough.

CovertJaguar:
Ok....this review system has weird limitations when submitting multiple reviews. -.-

src/main/java/mods/railcraft/common/blocks/logic/InventoryLogic.java
liach: Didn't notice that fact. My bad, fixed.

CovertJaguar: Are you confusing the tile/cart adapter with the inventory object?

src/main/java/mods/railcraft/common/util/inventory/IInventoryImplementor.java
liach: Added documentation to clarify as well.

Issue #1563, opened 2018-12-12
Question: Empty Lists in Config File


cpburnz:
This is probably a trivial question. How exactly to you format an empty list in the *railcraft.cfg*? In the version I'm running, 1.10.2-10.2.0, I want to remove the fuel cost from the world spikes. The config file says,

> if the list is empty, Worldspikes will not require fuel

How do you specify an empty list? Are these just multi-line strings, or proper lists? I'm not familiar with this file format and I can't find any information on it. It looks like a hybrid of JSON and INI, or similar to YAML. The file contains:

```
worldspikes {
    # ...
    fuel {
        # ...
        S:passive <
            railcraft:dust#0=2
            minecraft:ender_pearl=4
            railcraft:dust#6=8
            railcraft:dust#7=12
         >
        # ...
```

liach:
Do so:
```
S:passive <
>
```

On a side note, this cfg format is specified by forge.

cpburnz:
Thanks.

PR #1655, opened 2019-01-14
Fixes #1583, fixes #1599, fixes #1653, fixes #1613

### LINKED ISSUE #1567: Tank/Boiler GUI glitches: tanks and temperature scaling mismatchs.
**Description of the Bug**
In both tanks and boilers GUIs, the scale and the max of temperature, fluid amount, etc. are wrong.

**To Reproduce**
1. Build a HP steam boiler,
2. Make it heating,
3. Look the steam, water and temperature levels...

**Expected behavior**
The tank displays when the boiler is realy full, and not less-than-half full...
The temperature display doesn't get overlapped over some other textures...

**Screenshots & Video**
https://axeland.re/2A3h7d9
https://axeland.re/2A0lDcm
[Boiler Bug](https://www.youtube.com/watch?v=TTc39PlIci0)

**Environment**
 - Railcraft 12.0.0-alpha-9
 - Forge 14.23.5.2780
 - I don't thing it's a mod conflict...

**Additional context**
_None_

### LINKED ISSUE #1583: Firebox crash
**Description of the Bug**
This crash happens whenever I try to place down a solid/fluid fueled firebox
Crashlog:
https://gist.github.com/wormzjl/c40aba893d5fb66d06d2388565a4f83e

**Environment**
 - Version: railcraft-12.0.0-beta-1
 - Forge version. 2796
 - Possible conflicting mods and their versions.

### LINKED ISSUE #1599: [Suggestion] Config option for per-block capacity of the multiblock tank
Could be handy for modpack tweaking

### LINKED ISSUE #1613: [1.12.2] Fueled Boiler and Tank Bug
I want to report you more bug / problem on the latest version of Railcraft:
Config: Java 1.8.0_191 64 bit Forge 14.23.5.2779 (I also tried forge 14.23.5.2796, test done only solo and in creative

first of all the Fluid Fueled Boiler and Solid Fueled Boiler make crash minecraft of they are posed 

then I noticed several problems:
the water tank does not fill with water alone (which should normally be the case).
the iron and steel tanks have the same capacity of storage

the iron, steel and water tank do not display the encrypted capacity (the precise number of bucket).

Log : https://pastebin.com/j2nZ1D45

If need more info do not hesitate to contact me

### LINKED ISSUE #1653: Steel and iron tank colored valves do not have the correct texture in world
**Description of the Bug**
When a steel or iron tank valve is placed in world it does shows the default steel and iron textures respectively rather then the correct colored variant.

**To Reproduce**
1. Grab a colored steel or iron tank valve
2. Place it in world
3. Voila

**Expected behavior**
The tank valves show the correct color in world.

**Screenshots & Video**
Look up in the top left corner to see what they really are.

liach:
**The Issue**
 - #1583 
 - #1599 
 - ~#1567~ 
 - #1563 
 - #1653 
 
**The Proposal**
 - #1583: optional map to bypass npe
 - ~#1567: serialize capacity in standard tank~ Not a good resolution, leaving for further investigation
 - #1599: added config option
 - #1653: return base state instead of getting default state from block
 - #1613: first part is #1583, second part is #1567 
 
**Possible Side Effects**
 - Will affect if cj is already refactoring these multiblocks
 - Boiler still has some issues and needs deeper digging.
 
**Alternatives**
For the config option, I allow people to set a large value so that people won't bitch if they include rc in zimbabwe watt packs.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/multi/TileBoilerFirebox.java
+++ b/src/main/java/mods/railcraft/common/blocks/multi/TileBoilerFirebox.java
@@ -36,8 +36,8 @@ import net.minecraftforge.fluids.FluidStack;
 import net.minecraftforge.fml.relauncher.Side;
 import net.minecraftforge.fml.relauncher.SideOnly;
 import net.minecraftforge.items.CapabilityItemHandler;
+import org.jetbrains.annotations.Nullable;
 
-import javax.annotation.Nullable;
 import java.io.IOException;
 import java.util.Random;
 import java.util.function.Predicate;
@@ -57,7 +57,7 @@ public abstract class TileBoilerFirebox extends TileBoiler implements ISidedInve
     protected final TankManager tankManager = new TankManager();
     protected final FilteredTank tankWater = new FilteredTank(4 * FluidTools.BUCKET_VOLUME, this) {
         @Override
-        public int fillInternal(@org.jetbrains.annotations.Nullable FluidStack resource, boolean doFill) {
+        public int fillInternal(@Nullable FluidStack resource, boolean doFill) {
             if (!isValidMaster()) return 0;
             return super.fillInternal(onFillWater(resource), doFill);
         }
@@ -75,6 +75,9 @@ public abstract class TileBoilerFirebox extends TileBoiler implements ISidedInve
         tankManager.add(tankWater);
         tankManager.add(tankSteam);
 
+        tankWater.setCanDrain(false);
+        tankSteam.setCanFill(false);
+
         boiler = new SteamBoiler(tankWater, tankSteam);
         boiler.setTile(this);
     }
@@ -84,8 +87,8 @@ public abstract class TileBoilerFirebox extends TileBoiler implements ISidedInve
         int capacity = getNumTanks() * FluidTools.BUCKET_VOLUME;
         tankManager.setCapacity(TANK_STEAM, capacity * getSteamCapacityPerTank());
         tankManager.setCapacity(TANK_WATER, capacity * 4);
-        boiler.setMaxHeat(pattern.getAttachedData(BoilerData.EMPTY).maxHeat);
-        boiler.setTicksPerCycle(pattern.getAttachedData(BoilerData.EMPTY).ticksPerCycle);
+        boiler.setMaxHeat(pattern.getAttachedDataOr(BoilerData.EMPTY).maxHeat);
+        boiler.setTicksPerCycle(pattern.getAttachedDataOr(BoilerData.EMPTY).ticksPerCycle);
     }
 
     public boolean isBurning() {
@@ -325,9 +328,8 @@ public abstract class TileBoilerFirebox extends TileBoiler implements ISidedInve
         return capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY || super.hasCapability(capability, facing);
     }
 
-    @Nullable
     @Override
-    public <T> T getCapability(Capability<T> capability, @Nullable EnumFacing facing) {
+    public @Nullable <T> T getCapability(Capability<T> capability, @Nullable EnumFacing facing) {
         if (capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY) {
             return CapabilityItemHandler.ITEM_HANDLER_CAPABILITY.cast(ItemHandlerFactory.wrap(this, facing));
         }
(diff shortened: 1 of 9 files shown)


src/main/java/mods/railcraft/common/blocks/multi/TileBoilerFirebox.java
```diff
@@ -207,14 +210,14 @@ public void readFromNBT(NBTTagCompound data) {
     @Override
     public void writePacketData(RailcraftOutputStream data) throws IOException {
         super.writePacketData(data);
-//        tankManager.writePacketData(data);
+        tankManager.writePacketData(data);
```
CovertJaguar: This is only required in things that render the FluidStack in a TESR.

src/main/java/mods/railcraft/common/core/RailcraftConfig.java
```diff
@@ -250,6 +251,8 @@ private static void loadBlockTweaks() {
 
         allowTankStacking = get(CAT_TWEAKS_BLOCKS + ".irontank", "allow.stacking", true, "Change to '{t}=false' to disable the stacking of Iron Tanks");
 
+        tankPerBlockCapacity = get(CAT_TWEAKS_BLOCKS + ".irontank", "capacity.per.block", 1, 16 * FluidTools.BUCKET_VOLUME, 1600000, "Allows you to set how much milibucket of fluid each iron tank block can carry, min=1, default=16000, max=1600000");
```
CovertJaguar: This doesn't really need to be millibuckets, that will just confuse people. Buckets are fine grained enough.

src/main/java/mods/railcraft/common/fluids/tanks/StandardTank.java
```diff
@@ -189,4 +191,19 @@ public boolean isHidden() {
     public void setHidden(boolean hidden) {
         this.hidden = hidden;
     }
+
+    @Override
+    public StandardTank readFromNBT(NBTTagCompound nbt) {
+        super.readFromNBT(nbt);
+        if (nbt.hasKey("Capacity", Constants.NBT.TAG_INT))
+            capacity = nbt.getInteger("Capacity");
```
CovertJaguar: Do we really want to serialize this? Doesn't this get in the way of balance changes on our end?

src/main/java/mods/railcraft/common/fluids/TankManager.java
```diff
@@ -84,12 +84,14 @@ public void readTanksFromNBT(NBTTagCompound data) {
     public void writePacketData(RailcraftOutputStream data) throws IOException {
         for (StandardTank tank : tanks) {
             data.writeFluidStack(tank.getFluid());
+            data.writeInt(tank.getCapacity());
```
CovertJaguar: The client should always already know this.

CovertJaguar (changes requested):
I think the real problem is that the capacity calc should be in onPatternChanged instead of onPatternLocked. That _should_ fix it without requiring changes to TankManager and StandardTank.

src/main/java/mods/railcraft/common/fluids/TankManager.java
liach: this is set when the multiblock pattern change notification is received. seems it is no longer received on the client?

CovertJaguar: well there are three "pattern changed" event functions... onPatternChanged is called on the client... onPatternLocked, not so much.

StructureLogic fixes this mess.

CovertJaguar: Wait no.... onPatternChanged is gated by a isHost check....

CovertJaguar: I scrapped this entire chunk of code and started over for StructureLogic because it was so convoluted and inconsistent.

CovertJaguar: The Tanks and the Boilers are the only ones that use onPatternLock, just call it from the readData function.

liach: You can finish your code first and then I can rework on this. I can work on other issues, e.g. jei stuff

liach:
@covertjaguar I have dropped #1567 related change so that we can derive a better fix for that. The other bugs are clear and can be fixed as it is.

CovertJaguar (approved):

PR #1656, opened 2019-01-14
Fixes #1604, fixes #1643

### LINKED ISSUE #1604: Rolling machine recipes are not copied to crafting tables
**Description of the Bug**
When the factory module is disabled, rolling machine recipes are not copied to the crafting table.
~When disabeling the Factory module, things like tracks still use rails in their recipes. In this configuration the tracks would not be normally craftable. Also the rolling machine doesn't show recipes in JEI.~

**To Reproduce**
1. Disable Factory Module in Config
2. Look up recipes in JEI

**Expected behavior**
Rails, etc. craftable through crafting table.
~Alternative recipes that don't utilize items only craftable through the factory module.~

**Environment**
Railcraft 12.0.0-beta2
forge 14.23.2796

### LINKED ISSUE #1643: Do not register recipes if corresponding device is disabled
Like this
image

liach:
**The Issue**
 - #1604 failed to copy rolling machine recipes
 - #1643 hide jei recipe categories with no catalysts
 
**The Proposal**
 - Do not register disabled catalysts and hide categories manually.
 - Make sure the copied rolling machine recipes have distinct ids by making guessname return distinct names for item stacks.
 
**Possible Side Effects**
 - Consider guessname tweaks for other objects.
 - More postinit tweaks are a longer goal.
 
**Alternatives**
 - None aware.


src/main/java/mods/railcraft/common/plugins/forge/CraftingPlugin.java
```diff
@@ -206,7 +206,7 @@ public static ResourceLocation getNameFromOutput(ItemStack output) {
         if (input instanceof IForgeRegistryEntry) {
             return ((IForgeRegistryEntry) input).getRegistryName();
         } else if (input instanceof ItemStack) {
-            return ((ItemStack) input).getItem().getRegistryName();
+            return getNameFromOutput((ItemStack) input); // prevents duplicate names!
```
CovertJaguar: This seems incomplete? What about the other optional input types? They can still generate non-unique names.

I think perhaps the correct solution is to merge the two functions. Find a resource location, then pass it through the MultiSet unique counter to generate a unique one.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/plugins/forge/CraftingPlugin.java
liach: So converting all those elements to item stacks and make a name?

CovertJaguar: No need to convert them to ItemStacks, that is redundant. You already have a ResourceLocation, just add it to the recipeNames Multiset and append "$count" to the end of the path.

liach: Done.

CovertJaguar (approved):

liach:
eh, the travis pr is still in queue

Issue #1578, opened 2018-12-25
BootstrapMethodError


walker-dev:
**Description of the Bug**
crash on startup see log

**Environment**
 - Version railcraft-12.0.0-beta-1

**Additional context**
[Crash Log Here](https://gist.githubusercontent.com/boosterdev/13122d2c4293e463aa5d6b6f412dfd40/raw/90b1f00b54fc0339a3766a07f2ff1bf3170e5571/gistfile1.txt)

Ignore me for a few days and celebrate chrismas mate :)

Onigeshi:
Caused by: java.lang.BootstrapMethodError: java.lang.IllegalAccessError: no such constructor: mods.railcraft.common.items.ItemFilterBeeGenome.<init>()void/newInvokeSpecial

Install forestry, gendustry or binnies mods.. I don't know from which one of this is this Genome Filter but after installing all these 3 mods it didn't crash anymore..

Yzelast:
looks like installing forestry fix the crash...

navaroneee:
Is forestry needed?

liach:
Looks like we need to add another wrapping supplier to prevent class loading

navaroneee:
Ah ok, fine Merry xmas, and tnks for this gift :D

liach:
2f7415d10423702f3082079aa2c1e33c0dff685a

liach:
Is this fixed in beta 2?

walker-dev:
Yes it is fixed

Trinsdar:
this may have reappeared in beta 4.

liach:
@Trinsdar Log/crash report please.

Trinsdar:
https://paste.dimdev.org/dumebinomi.rb
I might have to test outside my dev instance just to make sure

liach:
aye, cannot see it in china, dunno why. if you can put it on pastebin or gist that is better

Trinsdar:
https://gist.github.com/Trinsdar/e90b133fff842eb0fae921b06e1d7749

liach:
hmm, i cannot find the cause this time. adding a mod prerequisite for this item might be the best solution, i guess.

3TUSK:
~~Successfully replicated this issue in a decompiled dev environment. All details are present in this crash report: https://paste.ubuntu.com/p/dgfCqBdssp/~~

Edit: this is exactly the reason why we have `deobfCompile`. Use that. Period.

liach:
[...]

PR #1660, opened 2019-01-18
Effect refactor

liach:
**The Issue**
 - Client and server (logical side) effects are called from the same class; people might miss which thread they are calling from, etc.
 - Trade station ignores careers. Added career support for trade stations.
 - Trade station recipe generation has some issues; only level 1
 - Chests have incorrect particles.
 - Trinsdar reports in #1578 that the error is happening again in beta 4.
 
**The Proposal**
 - Split effect into client and host effect classes.
 - add more optional method annotations to methods with mod classes in signature to prevent classloading
 - add career handling for trade station. another 2 buttons for career rotation.
 - I make trade stations generate recipes from nearby villagers when possible.
 - Add custom particles so that we can stitch the right texture to particles (block crack/dust)
 - Now trade station collect exp and increase reputation of its owner in the village when trading
 
**Possible Side Effects**
 - Might break existing api users that are unaware of contracts of our effect factory.
 - Changed default owner of rc tiles to unknown (for player reputation change when trading with trade station)
 
**Alternatives**
 - Move all effects to be sent on the host and received by client: good support for plugins, but it is way too much work


src/main/java/mods/railcraft/common/util/charge/ChargeNetwork.java
```diff
@@ -662,7 +663,7 @@ public void zap(Entity entity, Charge.DamageOrigin origin, float damage) {
                 }
                 if (remainingDamage > 0.1 && entity.attackEntityFrom(origin == Charge.DamageOrigin.BLOCK ? RailcraftDamageSource.ELECTRIC : RailcraftDamageSource.TRACK_ELECTRIC, remainingDamage)) {
                     removeCharge(chargeCost);
-                    Charge.effects().zapEffectDeath(entity.world, entity);
+                    HostEffects.INSTANCE.zapEffectDeath(entity.world, entity);
```
CovertJaguar: This one needs to remain unchanged, if only so we know that API is working.

CovertJaguar (changes requested):

CovertJaguar (approved):

src/main/java/mods/railcraft/common/util/charge/ChargeNetwork.java
liach: Done. However, api needs another merge as you merged that pr prematurely.

CovertJaguar (approved):

liach:
Should be ready. Wait for travis and do squash merge.

PR #1661, opened 2019-01-18
Avoid writing a whole game profile

liach:
See https://github.com/Railcraft/Railcraft-API/pull/29
(Small commit, done on github)
Split from #1655

**The Issue**
 - https://github.com/Railcraft/Railcraft-API/pull/29
 
**The Proposal**
 - Write a clean game profile instead.
 
**Possible Side Effects**
 - We don't use skin, so this should be fine.
 
**Alternatives**
 - None


liach:
Moreover, this is the only other place where we use `NBTUtil.writeGameProfile` besides that api one.

Issue #1607, opened 2018-12-29
[1.12.2] Sponge Forge server crash while using Manual Rolling Machine


054545641:
**Description of the Bug**
Producing (at least) something on Manual Rolling Machine will cause a server crash on SpongeForge Server. And there will be java.lang.IncompatibleClassChangeError in the Crash-report.

**To Reproduce**
1. Place a Manual Rolling Machine anywhere
2. Right click to open GUI of the Manual Rolling Machine
3. Put a Iron Ingot into it
4. Click Track Parts and wait for its completion

**Expected behavior**
A Track Part should be made normally without crash.

**Logs & Environment**
 - Environment:
Forge version 1.12.2-14.23.5.2779
Sponge version 1.12.2-2768-7.1.4
Railcraft version 12.0.0-beta-2
 - Full log: 
[debug.log](https://gist.github.com/054545641/13d7694f12bdfae396b699438bac889a)
[latest.log](https://pastebin.ubuntu.com/p/cj3yzHsrPx/)
 - Crash report (If available): 
[Crash-reports-server.txt](https://pastebin.ubuntu.com/p/zYfdkppZs9/)

liach:
Sponge bug.

liach:
@CovertJaguar Apparently sponge doesn't want our inventory to implement iterable for something other than `Iterable<org.spongepowered.api.item.inventory.Inventory>`. Blame Mumfrey.

CovertJaguar:
They make a change like that and expect it to break nothing? Lol.

Closing due to _Not Railcraft_

liach:
I think I will start working on the sponge compat soon, at least fixing this thing

PR #1663, opened 2019-01-18
Remove iterable from IInventoryComposite

liach:
This resolves the interface conflict with sponge inventory api 7 and is not too costly.

**The Issue**
 - #1607
 
**The Proposal**
 - Remove iterable from the `IInventoryComposite` interface.

**Why**
 - The iterable is not very widely used in Railcraft code (as seen in the changes here)
 - Sponge API is already finalized for Minecraft 1.12, making change on their side hard
 - Even if I use compat mod to make things run, the iterable contracts from two sides are different and will cause more subtle bugs. (Sponge contract: iterator iterates through rows; Railcraft: through sub inventory adaptors, which is actually the singleton inventory adaptor)
 
**Possible Side Effects**
 - This change should be clean and have no other side effects (code compiles)
 
**Alternatives**
 - Keeping as it is is not helpful given now the overwhelming majority of minecraft plugin-mod hybrid servers are based on SpongeForge. Though there is ftbutil, etc., they are more limited than sponge in general.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/logic/InventoryLogic.java
+++ b/src/main/java/mods/railcraft/common/blocks/logic/InventoryLogic.java
@@ -45,8 +45,8 @@ public abstract class InventoryLogic extends Logic implements IInventoryImplemen
     }
 
     @Override
-    public Iterator<InventoryAdaptor> iterator() {
-        return composite.iterator();
+    public Iterator<InventoryAdaptor> adaptors() {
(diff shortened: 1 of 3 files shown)


CovertJaguar (approved):

broncotc:
Thanks for fixing this, so happy that it's fixed after messing up with the sponge compatibility crash.

PR #1665, opened 2019-01-19
Fixes #1664

### LINKED ISSUE #1506: indev-1.12.2 Locomotive Will Not Reverse when Clicking R inside the locomotive Gui
indev-1.12.2 Locomotive Will Not Reverse when Clicking R inside the locomotive Gui

https://www.youtube.com/watch?v=YEP1rWlII5o

### LINKED ISSUE #1664: Update trackman's backpack to accept track kits
**Is your feature request related to a problem? Please describe.**
This feature request is to update the trackman's backpack to accept track kits since this is the preferred way of modding rails now. 

Currently, the track kits can't be placed int he trackman's backpack. It would be desired for it to be the default accepted (without custom config)

liach:
**The Issue**
 - #1664 
 - #1506
 - Inventory composite crashes on iron golems
 
**The Proposal**
 - Change the throw to an empty iterator since some entities don't have inventories
 - Cleans up some forestry related code
 - Syncs reversal status on gui return to help fix #1506 
 - Change track kit meta handling to take oredict wildcard as a special case for forestry backpacks
 
**Possible Side Effects**
 - The oredict wildcard special case might affect other usages which I am not sure.
 - Dunno why you decided to throw originally for the inventory composite.
 
**Alternatives**
 - I could have returned true meta, but I did not because I don't know if that hurts other usages of track kit

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/tracks/outfitted/ItemTrackKit.java
+++ b/src/main/java/mods/railcraft/common/blocks/tracks/outfitted/ItemTrackKit.java
@@ -42,6 +42,7 @@ import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
 import net.minecraftforge.fml.relauncher.Side;
 import net.minecraftforge.fml.relauncher.SideOnly;
+import net.minecraftforge.oredict.OreDictionary;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


src/main/java/mods/railcraft/common/util/inventory/IInventoryComposite.java
```diff
@@ -36,8 +33,8 @@
 
     default Iterator<InventoryAdaptor> adaptors() {
         return InventoryAdaptor.of(this)
-                .map(Iterators::singletonIterator)
-                .orElseThrow(UnsupportedOperationException::new);
+                .<Iterator<InventoryAdaptor>>map(Iterators::singletonIterator)
+                .orElseGet(Collections::emptyIterator);
```
CovertJaguar: You have misidentified the problem. InventoryComposite, the standard implementation of this interface, needs to override these new functions.

I'll fix it, so revert this.

CovertJaguar (changes requested):

liach:
i will definitely revert that as you initially designed this system, which i do not yet understand throughoutly.

liach:
Reverted that change.

CovertJaguar (approved):

Issue #1558, opened 2018-12-07
Increase void chest item deletion rate


liach:
**Is your feature request related to a problem? Please describe.**
Void chest only destroys one item in a stack each second. This makes it extremely useless compared to sheer lava burning or cacti. People have avoided lava or cacti for lag, but void chest does not appear to be better.

**Describe the solution you'd like**
Allow void chest to destroy at faster rates based on how full it is.

**Describe alternatives you've considered**
Keep it as it is. Users have complained that this chest is stupid for this weakness.

**Additional context**
None.

CovertJaguar:
I'm fine with making it based on fill percentage. I made it slow originally to possibly give you time to grab anything you mistakenly shift click into the chest.

Issue #291, opened 2014-09-14
Metals chest won't touch ingots that are registered in the ore dict but not used by Railcraft 


l3CY:
Love the idea of the metal chest, but there a big problem.  It won't touch ingots that Railcraft itself doesn't utilize in crafting.  As a result it has very limited functionality when used with other mods.  Any chance that its usability could be expanded to condense (attempt a 3x3 crafting operation) with anything placed inside that's registered as an ingot?

mikeloeven:
Whats the status of this issue currently on FTB 1.11.2 and the metals chest is still not compressing Ic2 and Tinkers ingots

bookerthegeek:
Because ic2 refined iron and railcraft steel is dictionary together.

Generalcamo:
Is this still valid? I don't think the Metals Chest exists anymore.

bookerthegeek:
@Generalcamo Dont speak such evil!

CovertJaguar:
Code is still there. Just no rendering and need to split to own block.

liach:
Railcraft added quite a lot of metals. Now it can handle bronze invar or brass.

Generalcamo:
Irrelevant, there are far more metals added by mods like EnderIO, Thermal Foundation, TechReborn, and more.

CovertJaguar:
You'd need some way to look up crafting recipes as we can't just assume that you can craft nine ingots into one block.

PR #1669, opened 2019-01-20
Fixes 2 chest issues

liach:
**The Issue**
 - #1558
 - #291
 
**The Proposal**
 - Void chests now delete based on fullness.
 - Metals chest now convert by recipes.
 - Increase compression rate to be same as the void chest deletion rate (the old rate is too slow given the chest's expensive recipe)
 
**Possible Side Effects**
 - Metals chest no longer convert other mods' items to railcraft's items (might not be bad as dog poo cookies might happen)
 
**Alternatives**
 - Deleting void chest contents too fast may cause bad results when your hand slip.


CovertJaguar:
I'll be honest, the reason I sat on this is because one of the original reasons I implemented the metals chest was as a response to the unification problem.

Removing that feature from it pretty much removes its entire reason for existence.

liach:
i will add a converter to convert oredicts to rc form. the converter will run before the condensation. is that good enough?

liach:
Closing. Do not delete the branch though; I will cherry pick metals chest commit later.

Issue #1689, opened 2019-01-29
Cubic Chunks Compatibility


TheBigBadWolf2:
[2019-01-29-3.log.gz](https://github.com/Railcraft/Railcraft/files/2809594/2019-01-29-3.log.gz)
Using Cubic Chunks, when placing a certain blocks below y=0, the console gets spammed, and GUIs for blocks will not open.
Forge Version: 14.23.5.2779
Rail Craft Version: 10.0.0-beta-4
Cubic Chunks Version: 0.0.912.0
Cubic World Gen Version:  0.0.29.0

liach:
I have told people in discord repeatedly that railcraft depends on a negative y value for opening Minecart guis. I am actually fairly suprised that the game didn't crash straight. I'd like @barteks2x to find a viable solution instead.

Niko-sk2x:
Yes you can fix it. Whether someone wants to fix it, is a different matter.

It would involve around 10 lines of changes. First is just remove that log line that shows the spammy output for things below y=0.
The other is to serialize BlockPos to 3 ints instead of a single long (afaik, the GUI issues are only beyond -2048 to 2074 height range).

At least this should fix the 2 issues mentioned here. And I haven't found any other obvious issues that would apply to cubic chunks.

liach:
But vanilla gui handler only support 3 ints. For cubicchunks, all 3 of them will be used for blockpos; as a result, railcraft cannot open minecart guis.

For the 3 int and long thing, Railcraft should be saving 3 ints mostly. However, it is not a long-term solution. In 1.13+, Mojang uses the long serialization vastly, making it a pain for cubic chunks.

Niko-sk2x:
The X and Z coordinates have 6 bits of unused space each. Or, since it appears to be used as a flag, you could just pass Integer.MIN_VALUE and test for that.

For 1.13/1.14+ I'm probably going to change the way longs are used for block pos to have the user choose a tradeoff between horizontal and vertical size. And only optionally patch the long usages. And as far as I'm aware, this is only done for some deep internals that modders should rarely directly touch.
And since there is still 4096 blocks of usable space, it's not too big of an issue, but it would still be great if it was changed, as this is just a few additional lines of code (here is where I wish what PacketBuffer does would be more available to mods, as that is something I can patch)

CovertJaguar:
[...]

PR #1704, opened 2019-02-10
Remove or deprecate y<0 statements for cubic chunks

liach:
**The Issue**
 - #1689
 - Removed y<0 checks
 
**The Proposal**
 - Remove y<0 checks
 - Use integer min value to indicate entity gui instead
 
**Possible Side Effects**
 - Might break some safety guarantees; not sure
 
**Alternatives**
 - Some people use cubic chunks, so adding its compatibility is a nice choice. The change is not too large, either.

@Barteks2x please look at this as well.

### DIFF
--- a/src/main/java/mods/railcraft/common/gui/GuiHandler.java
+++ b/src/main/java/mods/railcraft/common/gui/GuiHandler.java
@@ -42,7 +42,7 @@ public class GuiHandler implements IGuiHandler {
     public static void openGui(EnumGui gui, EntityPlayer player, World world, Entity entity) {
         if (Game.isHost(world)) {
             if (gui.hasContainer()) {
-                player.openGui(Railcraft.getMod(), gui.ordinal(), world, entity.getEntityId(), -1, 0);
+                player.openGui(Railcraft.getMod(), gui.ordinal(), world, entity.getEntityId(), Integer.MIN_VALUE, 0);
             }
         } else if (!gui.hasContainer()) {
             FMLClientHandler.instance().displayGuiScreen(player, FactoryGui.build(gui, player.inventory, entity, world, entity.getEntityId(), -1, 0));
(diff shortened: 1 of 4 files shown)


Niko-sk2x:
Looks good.

The only thing that it may break is `setBlockState`, `getLightFor`, `setLightFor` and `getLightSubtracted` accessed **directly on `Chunk`** (all others, including `getBlockState` should be fine). If there are any places where this is done (unlikely), I would recommend using `World.isOutsideBuildHeight`.

src/main/java/mods/railcraft/common/gui/GuiHandler.java
```diff
@@ -42,7 +42,7 @@ public static void openGui(EnumGui gui, EntityPlayer player, World world, int x,
     public static void openGui(EnumGui gui, EntityPlayer player, World world, Entity entity) {
         if (Game.isHost(world)) {
             if (gui.hasContainer()) {
-                player.openGui(Railcraft.getMod(), gui.ordinal(), world, entity.getEntityId(), -1, 0);
+                player.openGui(Railcraft.getMod(), gui.ordinal(), world, entity.getEntityId(), Integer.MIN_VALUE, 0);
```
CovertJaguar, line 45: This probably works, but if it doesn't we could always pack it into to gui.ordinal() parameter. That's how Forestry sends page numbers for its GUIs.

CovertJaguar (approved):

src/main/java/mods/railcraft/common/gui/GuiHandler.java
liach: Uhh, using an `~` bitwise operation may be too confusing for coders compared to `-`, though `-` is an incorrect operation.

src/main/java/mods/railcraft/common/gui/GuiHandler.java
```diff
@@ -42,7 +42,7 @@ public static void openGui(EnumGui gui, EntityPlayer player, World world, int x,
...
             }
         } else if (!gui.hasContainer()) {
             FMLClientHandler.instance().displayGuiScreen(player, FactoryGui.build(gui, player.inventory, entity, world, entity.getEntityId(), -1, 0));
```
CovertJaguar, line 48: What about this line? I think it might only affect the TNT carts.

PR #1708, opened 2019-02-11
Fixes #1623

### LINKED ISSUE #1623: Brass recipe does not work when inventory tweaks is loaded
**Description of the Bug**
If inventory tweaks is loaded brass cannot be crafted from 3 copper ingots and one zinc ingot

**To Reproduce**
1. install inventory tweaks
2. get three copper ingots and one zinc ingot
3. try to craft brass

**Expected behavior**
brass should craft from 3 copper ingots and one zinc ingot

**Screenshots & Video**

**Logs & Environment**
latest.log: https://paste.dimdev.org/qamoluhafi.md

**Additional context**
Railcraft 12.0.0-beta2
forge14.23.5.2779
inventory tweaks-1.63
Edit: tested with latest alpha version of inventroy tweaks on curseforge and still happens

liach:
**The Issue**
 - #1623 
 - When an empty oredict entry exist, we assume it has items and withdraw our recipes.
 
**The Proposal**
 - Change the check to test against the existence of ores.
 - When empty oredicts for bronze/invar dust is added, bronze/invar recipe breaks as well, so this method.
 
**Possible Side Effects**
 - This method is only used in ingot recipe registration of 3 alloys and in `isOreType` in oredict plugin. Does not produce exceptional behavior as I can see.
 
**Alternatives**
 - Removing the oredict check for brass can only help fixing brass, but cannot work for bronze or invar when invtweak is present and no other invar/bronze mod is present.

### DIFF
--- a/src/main/java/mods/railcraft/common/plugins/forge/OreDictPlugin.java
+++ b/src/main/java/mods/railcraft/common/plugins/forge/OreDictPlugin.java
@@ -76,7 +76,7 @@ public final class OreDictPlugin {
     }
 
     public static boolean oreExists(String name) {
-        return OreDictionary.doesOreNameExist(name);
+        return !OreDictionary.getOres(name, false).isEmpty();
     }
 
     public static Set<IBlockState> getOreBlockStates() {


CovertJaguar:
It should be noted that this is still dependent upon the mod load order. It is not safe to use in pre-init at least.

liach:
`OreDictPlugin.isOreType` is not used in preinit as well; it is used in anvil repair checks and bore mining checks (during mining)

Issue #1718, opened 2019-02-15
major infinite steel exploit


EthernalVoid:
* **Modpack Version**: 1.0.1
* **Issue**: You can make infinite steel by smelting iron tank walls/gauge/valve [railcraft] in a blast furnace [from Railcraft]
* **Is it Repeatable?**: yes

> 4x _iron plates = 8x iron tank walls
> smelting 1 iron iron tank wall in a blast furnace gives _4 STEEL_ ingots back

> so 32 steel in total from 4ish iron

* **Mod/s Affected**: railcraft
* **Known Fix**: fix recipes

liach:
:thinking: Probably wanted to give 4 nuggets initially. good catch!

liach:
Uh yeah, it was indeed accidentally messed up during code maintenance 

https://github.com/Railcraft/Railcraft/blob/b7ba6bea1e52ac55792377a1336497c07282ed18/src/main/java/mods/railcraft/common/blocks/multi/TankDefinition.java#L63

Should be easy to fix.

EthernalVoid:
Superbly quick response, much obliged :)

CovertJaguar:
Oh... the tank recipe produces multiple product. I didn't catch that.

PR #1723, opened 2019-02-18
Fix entity gui and some assorted stuff

liach:
**The Issue**
 - Wrong GUI code might have broken tnt carts.
 - #1718 
 - Advancements aren't trigger for strucutre logics.
 - #1567
 
**The Proposal**
 - Add hooks for structure logic
 - Fix `-1`
 - Write capacity in packet sync
 - fix a minor issue in depth detection 

**Possible Side Effects**
 - Haven't seen any yet.
 
**Alternatives**
 - Just tried to do minimal change for these bugs.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/multi/TankDefinition.java
+++ b/src/main/java/mods/railcraft/common/blocks/multi/TankDefinition.java
@@ -61,7 +61,7 @@ public enum TankDefinition {
             Crafters.blastFurnace().newRecipe(block)
                     .name("railcraft:smelt_iron_tank")
                     .time(IBlastFurnaceCrafter.SMELT_TIME * 4)
-                    .output(RailcraftItems.INGOT.getStack(4, Metal.STEEL)).slagOutput(4)
+                    .output(RailcraftItems.NUGGET.getStack(4, Metal.STEEL)).slagOutput(1)
                     .register();
         }
     },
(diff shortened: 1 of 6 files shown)


src/main/java/mods/railcraft/common/blocks/multi/TankDefinition.java
```diff
@@ -61,7 +61,7 @@ public void defineRecipes(Block block) {
             Crafters.blastFurnace().newRecipe(block)
                     .name("railcraft:smelt_iron_tank")
                     .time(IBlastFurnaceCrafter.SMELT_TIME * 4)
-                    .output(RailcraftItems.INGOT.getStack(4, Metal.STEEL)).slagOutput(4)
+                    .output(RailcraftItems.NUGGET.getStack(4, Metal.STEEL)).slagOutput(1)
```
CovertJaguar, line 64: Why? Didn't I just revert this change? Almost all the other smelting recipes are 1:1 on the iron:steel ratio.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/blocks/multi/TankDefinition.java
liach: this is for #1718. Four iron plates can craft 8 tanks walls.

CovertJaguar: Gotcha!

CovertJaguar (approved):

Issue #1720, opened 2019-02-16
Detector Track Kits don't interact with Comparators


AzazelHD:
I had a hopper wagon, and when he is above a hopper it does not send any redstone signal to the comparator next to the hopper (it sends a redstone signal of power 1 = not enough to stop the wagon until its hopper is empty).

I don't know why you changed the original code... This mod literally fucked up all the track system in minecraft.

Could you fix this please?

I can't do a recolector because it doesn't work properly. Wagon does not stop over the hopper until it is empty

liach:
Lol comparator outputs for minecarts are passed with detectors. Redstone comparators do not detect minecarts.

AzazelHD:
I mean, the redstone track over the hopper, and the hopper near the comparator...

liach:
Hmm, need to check if this is a design issue of railcraft. try replacing the comparator with a redstone wire and see the output strength.

CovertJaguar:
Some pictures would help.

liach:
hmm, the comparator output does not update with the cart's inventory change.

CovertJaguar:
What is it attached to? The hopper?

CovertJaguar:
The issue here is that Detector TrackKits don't implement the Comparator integration. This is an oversight on our part.

PR #1725, opened 2019-02-18
Fixes comparator output of detector kit

liach:
**The Issue**
 - #1720 comparators not updated
 - Also fix command block cart detection.
 
**The Proposal**
 - Because a cart's inv may change while on track, the output needs to be constantly updated.
 
**Possible Side Effects**
 - I am unaware of any negative ones.
 
**Alternatives**
None.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/tracks/outfitted/kits/TrackKitDetector.java
+++ b/src/main/java/mods/railcraft/common/blocks/tracks/outfitted/kits/TrackKitDetector.java
@@ -99,6 +99,7 @@ public class TrackKitDetector extends TrackKitRailcraft implements ITrackKitEmit
     protected void setTrackPowering() {
         boolean notify = delay == 0;
         delay = CartConstants.DETECTED_POWER_OUTPUT_FADE;
+        theWorldAsserted().updateComparatorOutputLevel(getPos(), getTile().getBlockType());
         if (notify) {
             notifyNeighbors();
         }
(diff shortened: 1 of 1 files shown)


CovertJaguar (approved):

AzazelHD:
Sorry I am lost. You moved my issue to this threat?

liach:
what? threat? i don't know what you are thinking, but i've fixed the bug you reported.

AzazelHD:
I Opened another issue with comparators not sending redstone signal and it was closed :P

liach:
Uh? I closed "another issue" #1720 because I fixed it. Comparators are sending redstone signal now; they output the inventory status of the hopper cart, so if your hopper cart is empty, there is no redstone signal sent.

AzazelHD:
Yeah that one.
So if the hopper cart has items and under a redstone track there is another hopper, will it work?

Wait, i paste a video and i tell u the minute:
Min 23

https://www.youtube.com/watch?v=cSnogMr_ZBU&t=1444s 

Is that possible?

liach:
The setup in the video is getting comparator output from the hopper block instead of the track. Which block the comparator is detecting depends on the y-level of the comparator: it will detect the block at the same y-level as itself.

AzazelHD:
With railcraft that is possible? The built on the video. Because with the version I had the redstone power signal was of 1 instead of 15

liach:
Yes, but I recommend trying locking track kit and item loader instead. For the locking track kit, you should change it to holding mode with a crowbar.

AzazelHD:
Okey, tyvm :D 

And where is this new version to download?

liach:
wait for beta 6.

AzazelHD:
ah okey. Do you know when will it be released?

liach:
There is no estimated time of arrival. For more questions, please go to https://discordapp.com/invite/VyaUt2r as this is not a good thread (this is a pull request instead of an issue). Locking.

Issue #1716, opened 2019-02-15
Track Layer Cart is incapable of retrieving  tracts from adjacent chest carts


mariojuggernaut22:
**Description of the Bug**

**To Reproduce**

1. Place block/cart '...'
3. See '...'

**Expected behavior**

**Screenshots & Video**

**Logs & Environment**

 - Full log: 

 - Crash report (If available): 

**Additional context**

mariojuggernaut22:
**Description:**
Whenever you connect a Chest Cart to  the track layer cart, it fails to extract the  rail contents of the  box.
**To reproduce:**
Place a track  layer cart and chest cart together, join them  with he crowbar,  fill the chest with the tracks you need to be placed and set the layer to dispense those tracks.

liach:
Might be some issue in the maintenence cart codebase.

PR #1726, opened 2019-02-18
Fix track layer stock pulling

liach:
**The Issue**
 - Fix track layer stock pulling from chest carts #1716
 
**The Proposal**
 - Always pull; not waiting till the stock is empty. Also fixed null usage for item stacks.
 
**Possible Side Effects**
 - Might affect other maintenance carts as well (heard relayer had a similar bug)
 
**Alternatives**
 - I think this is the most obvious place and most correct place to fix.

### DIFF
--- a/src/main/java/mods/railcraft/common/util/inventory/InvTools.java
+++ b/src/main/java/mods/railcraft/common/util/inventory/InvTools.java
@@ -507,4 +507,16 @@ public abstract class InvTools {
     public static double calculateFullness(IInventoryManipulator manipulator) {
         return manipulator.streamSlots().mapToDouble(slot -> slot.getStack().getCount() / (double) slot.getMaxStackSize()).average().orElse(0.0);
     }
+
+    /**
+     * Checks if a stack can have more items filled in.
+     *
+     * <p>Callers: Be warned that you need to check slot stack limit as well!
+     *
+     * @param stack the stack to check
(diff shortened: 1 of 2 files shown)


src/main/java/mods/railcraft/common/util/inventory/InvTools.java
```diff
@@ -507,4 +507,16 @@ public static IBlockState getBlockStateFromStack(ItemStack stack) {
...
+    /**
+     * Checks if a stack can have more items filled in.
+     *
+     * <p>Callers: Be warned that you need to check slot stack limit as well!
+     *
+     * @param stack the stack to check
+     * @return whether the stack needs filling
+     */
+    public static boolean isStackFull(ItemStack stack) {
+        return stack.isEmpty() || stack.getCount() == stack.getMaxStackSize();
```
CovertJaguar: An empty stack is considered full? If this is intentional it should probably be documented in the javadoc.

Or is that supposed to be a less than?

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/util/inventory/InvTools.java
liach: Hmm, I initially meant to be "can fill". Yes, I should change that.

CovertJaguar (approved):

PR #1733, opened 2019-02-19
Make sure removed trains are actually removed in file

### LINKED ISSUE #1724: Train count grows exponentially and using up the RAM of the server
**Description of the Bug**
(Edited by @liach)
We are running your beta 5 version for 1.12 and when the world goes through a save process, in order to have a backup incase of a crash, Railcraft is trying to save all the locations of minecarts in the world, but during the save, something goes wrong, it duplicates the numbers creating ghost carts, and then tries to save again less than 30 seconds later, creating a constant double save loop, but with increasing file size and cart value, up until it goes over 20 GB of file size, and overpowers the ram on the server, usually every 45 minutes after a full reboot. Its also not overwriting the previous save file, but instead creating a new one, my guess because it thinks there are more minecarts than there actually are. We traced down the issue very carefully, its how we know where that issue is coming from. It usually reaches critical mass around 2,000 carts just before the server completely crashes and requires a reboot of the minecraft server.

(Original content below)

Attempting to disable anything in order to reduce or remove the constant saving issue of Ghost Mine Carts just sets up the client to constantly crash. 
**To Reproduce**

1. Place block/cart '...' Go into the mod config settings and disable anything at all. I tested each option. 
3. See '...'

**Expected behavior**

It completely crashes Minecraft halfway through load up
**Screenshots & Video**

https://pastebin.com/UXR9n0S5
**Logs & Environment**

 - Full log: 

 - Crash report (If available): 

https://pastebin.com/UXR9n0S5
**Additional context**

liach:
**The Issue**
 - Fixes #1724 
 - Existing servers can delete old `railcraft.trains.dat` file in `data` folder of world directory to relieve this.
 
**The Proposal**
 - Scan through all trains less frequently.
 - Make sure to mark the data as dirty after trains are removed.
 - Adds a utility to see if there is any duplicate cart in railcraft.trains.dat file.
 
**Possible Side Effects**
 - Not much I can see.
 
**Alternatives**
?

### DIFF
--- a/src/main/java/mods/railcraft/common/carts/Train.java
+++ b/src/main/java/mods/railcraft/common/carts/Train.java
@@ -83,10 +83,15 @@ public final class Train implements Iterable<EntityMinecart> {
 
     public static Object getTicker() {
         return new Object() {
+
+            int counter = 0;
+
             @SubscribeEvent
             public void tick(TickEvent.WorldTickEvent event) {
                 if (event.side == Side.SERVER && event.phase == TickEvent.Phase.END) {
-                    getManager(event.world).ifPresent(Manager::tick);
+                    counter++;
+                    if (counter % 32 == 0)
+                        getManager(event.world).ifPresent(Manager::tick);
                 }
             }
         };
@@ -494,6 +499,7 @@ public final class Train implements Iterable<EntityMinecart> {
                 if (train.isDead || train.isInvalid()) {
                     train.clear();
                     it.remove();
+                    data.markDirty();
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


CovertJaguar:
While an improvement, I doubt this has much affect on the underlying issue. The save data should be marked dirty eventually, or it would never save at all. At which point it should be purged of dead trains. If infinite trains are being generated I don't see how this could fix it.

liach:
But once a previously deleted train skips a save, the train can remain forever in the file. That might be?

CovertJaguar:
Not if the NBT tree is regenerated from the Train map held in Train.Manager on every save.

Issue #983, opened 2016-11-01
Village Chests Have No Loot


Zalaniar:
The chests inside the Railcraft buildings generate with nothing inside them. The server console gives this warning upon launch: 
`Server thread/WARN Couldn't find resource table railcraft:chests/village_workshop`
The chests are there, and other features of the buildings generate properly, but the chests are empty. Other village chests also generate normally.

CovertJaguar:
Sadly, my loot code is utterly broken and it has been pretty low priority to fix. At some point I'll probably need to bug @mezz to see how he did it in Forestry.

mezz:
Forestry's loot handling is a crazy hack...

I wrote a set of loot tables as if they were the vanilla loot tables, 
https://github.com/ForestryMC/ForestryMC/tree/mc-1.10/src/main/resources/assets/forestry/loot_tables/chests

and the load them  with my own loader:
https://github.com/ForestryMC/ForestryMC/blob/mc-1.10/src/main/java/forestry/core/loot/LootTableLoader.java

Then I inject my table's stuff into the matching vanilla one on this event:
https://github.com/ForestryMC/ForestryMC/blob/mc-1.10/src/main/java/forestry/core/EventHandlerCore.java#L101-L117

This is the only way I could come up with to do it without writing everything in code, which is poorly supported.

CovertJaguar:
My eyes burn! Its worse than I thought!

Ristellise:
lol. the sad reality of forestry mod loot tables...

brisingraerowing:
Some mods I use are using the [Additional Loot Tables Mod](https://minecraft.curseforge.com/projects/additional-loot-tables) to handle adding loot tables to Minecraft. Might be worth a look to see if it would work for you, or if you can glean some insight on how to get it to work.

mezz:
A recent Forge PR has allowed me to implement this in a less eye-burning way in 1.11:
https://github.com/ForestryMC/ForestryMC/commit/3e175aa8e6ebb10f213581ca6a80b741b00082a2

CovertJaguar:
@mezz my main issue with the current implementation is that it chokes hard on missing items, rather than ignore them. Which makes it a pain for people who disable Railcraft items or modules, they also have to provide their own loot tables with the disabled items removed.

mezz:
I should split up the loot table additions so they are one file per module to help with that.
Did you find a better way to do it in Railcraft or does it still have the same issue?

[...]

PR #1754, opened 2019-03-06
Fix loot table not customizable

### LINKED ISSUE #51: Railcraft loot cannot be removed item at a time, but pools can be removed
**Description:**
Railcraft (in quite buggy beta) doesn't allow removing specific items. Thankfully removing entire pools is doable

**Steps to reproduce:**

**Version Info (Exact versions only):**  
LootTweaker:  1.12.2-0.0.8
CraftTweaker/Minetweaker:  1.124.1.15
Forge:  14.23.5.2815
Minecraft:  1.12.2

**Logs, scripts and other useful information:**  
https://github.com/Railcraft/Railcraft/issues/1659

### LINKED ISSUE #1659: [Suggestion] Option to disable dungeon loot
**Is your feature request related to a problem? Please describe.**

A majority of the ingots, plates and gears added to dungeon loot have no use in my pack. Less loved metals like lead and aluminum were removed, and recipes that used them were changed.

Due to a myriad of issues with Railcraft, I have to leave the ingots, plates and gears enabled in the config. These items are then added to dungeon loot, which I don't like.

Add a config option to stop Railcraft from adding loot to vanilla loot tables. The Railcraft village structure can be unaffected by this, since removing Railcraft items would leave it empty. It's mostly rails anyway. 

**Describe alternatives you've considered**
I've tried using LootTweaker to remove the items, but to be honest it's way over my head.

Disabling plates, ingots or gears in the config causes certain trades to disappear, and makes the Railcraft village chest generate empty.

Disabling the entire world generation module would probably work, but this would also remove abyssal and quarried stone which are simply too beautiful to eliminate. 

**Additional context**

liach:
**The Issue**
 - Fix loot table not customizable, close #1659
 - Describe the problem if an issue doesn't exist. Be as clear as possible.
 - Also resolves Daomephsta/LootTweaker#51
 
**The Proposal**
 - Use vanilla loot table loading so that we can take account of customized loot tables in data folder and other mods, etc.
 - Now can just disable the table in loottweaker
 
**Possible Side Effects**
 - The old condition is removed because vanilla just fail fast when encountering and item that does not exist (without checking any condition first, unlike recipes)
 
**Alternatives**
 - This is a minimal change. Adding too much config or details for loot tables is too much and might cause more unexpected issues.

### DIFF
--- a/src/main/java/mods/railcraft/common/plugins/forge/LootPlugin.java
+++ b/src/main/java/mods/railcraft/common/plugins/forge/LootPlugin.java
@@ -10,30 +10,14 @@
 
 package mods.railcraft.common.plugins.forge;
 
-import com.google.common.base.Charsets;
-import com.google.common.io.Resources;
-import com.google.gson.*;
 import mods.railcraft.api.core.RailcraftConstantsAPI;
 import mods.railcraft.common.core.RailcraftConstants;
-import mods.railcraft.common.modules.RailcraftModuleManager;
-import mods.railcraft.common.util.misc.Game;
 import net.minecraft.util.ResourceLocation;
-import net.minecraft.world.storage.loot.*;
-import net.minecraft.world.storage.loot.conditions.LootCondition;
-import net.minecraft.world.storage.loot.conditions.LootConditionManager;
-import net.minecraft.world.storage.loot.functions.LootFunction;
-import net.minecraft.world.storage.loot.functions.LootFunctionManager;
-import net.minecraftforge.common.ForgeHooks;
+import net.minecraft.world.storage.loot.LootPool;
+import net.minecraft.world.storage.loot.LootTable;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.event.LootTableLoadEvent;
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent;
-import net.minecraftforge.fml.common.registry.ForgeRegistries;
-import org.apache.logging.log4j.Level;
-import org.jetbrains.annotations.Nullable;
-
-import java.io.IOException;
-import java.net.URL;
-import java.util.Random;
 
 import static net.minecraft.world.storage.loot.LootTableList.register;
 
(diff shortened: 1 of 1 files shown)


CovertJaguar:
The main problem I see here is that this removes the ability to add new pools to vanilla loot tables.

You remove the LootTableLoader, but it should probably be replaced with the updated version from Forestry described here:
Village Chests Have No Loot https://github.com/Railcraft/Railcraft/issues/983
https://github.com/ForestryMC/ForestryMC/commit/3e175aa8e6ebb10f213581ca6a80b741b00082a2

liach:
Hmm? We are still adding loot pools to vanilla loot tables.
https://github.com/Railcraft/Railcraft/pull/1754/files#diff-9eceae9539b15a958721a4942944adf6L88
Moreover, I do not believe that we need to do url checking like forestry does. It has no real point, as loot tables can be loaded from data folder in world saves.

CovertJaguar:
Yes we are. We only add village_workshop, the other five tables are vanilla defined, and without this code I believe they get ignored.

I don't know what you mean by URL checking, the code I linked takes the vanilla definitions and appends our pools to them.

I don't understand what tables in the data folder has to do with anything.

liach:
hmm? we still have the ability to add new pools to vanilla pools.

CovertJaguar:
Well at the time I wrote (or borrowed) this code, that wasn't a feature of Forge or the base game. The pools are defined currently in the Railcraft namespace. By what mechanic are they moved to the Minecraft namespace? That's what the code you removed was doing.

CovertJaguar:
Hmm... I may have falsely assumed the code you removed was doing more than just loading the json files.

CovertJaguar:
Now, onto the rest of the changes, you remove the ItemExistsCondition. 

This was causing the integration issues with LootTweaker?

What is the vanilla behavior when encountering an invalid item? I seem to recall it invalidated the entire loot table, not just the current pool, but my memory might be faulty.

liach:
https://github.com/Railcraft/Railcraft/pull/1754/files#diff-9eceae9539b15a958721a4942944adf6L94

It was broken because vanilla invalidates the whole loot table when encountering an invalid item; the encounter is made before any condition is processed by the loot table loader; thus, a condition cannot prevent the invalidation.

CovertJaguar:
Ah so the condition was pointless then.

Additionally, does per world loot table override Railcraft's loot pool additions or do they get added always? This seems to be issue #1659 and I don't see how these changes address that.

liach:
If the per-world loot tables have blanked out Railcraft tables that contains pools that are supposed to be added to vanilla tables, the Railcraft's loot pool additions will be removed. So for loottweaker users, they just need to clear out those railcraft tables, e.g. `railcraft:chests/simple_dungeon`.

Issue #1751, opened 2019-03-04
Option to disable "Saving trains" console spam


TheDeviantCrafter:
**Is your feature request related to a problem? Please describe.**
Railcraft spams the console whenever it saves the trains. This doesn't cause any noticeable lag and isn't useful to the average player.

**Describe the solution you'd like**
A config option to disable the message from appearing in console. By default, it shouldn't appear in console. This seems like debug information that should only be shown if you explicitly ask for it.

**Describe alternatives you've considered**
A mod like Console Filter could be used, but it's probably not worth the effort.

**Additional context**

liach:
Sure. Will make that log message debug level instead.

PR #1755, opened 2019-03-06
Make train saving message debug level instead

liach:
**The Issue**
 - #1751
 
**The Proposal**
 - Reduce console spam and irrational concerns.
 
**Possible Side Effects**
~
 
**Alternatives**
~


liach:
Build errored because of temporary progwml6 maven outage. Will rebuild later as player site is down now.

PR #1761, opened 2019-03-06
Rich biome works again; implement diffuse gen.

### LINKED ISSUE #1710: Rich biome settings for mines
**Description of the Bug**
Rich biome settings for mines don't work.

**To Reproduce**

1. Travel to mesa
2. Look at xray results

**Expected behavior**
Should have more ore and a higher density

**Analysis**
The white black list is broken.

**Additional context**
Left here so we do not forget about this bug.

liach:
**The Issue**
 - Close #1710
 - Also implemented diffuse gen with vanilla ore gen
 
**The Proposal**
 - Tweaked logic to make sure black white lists work.
 
**Possible Side Effects**
 - The white black list should be right this time as we have tests.
 
**Alternatives**

### DIFF
--- a/src/main/java/mods/railcraft/common/util/misc/WhiteBlackList.java
+++ b/src/main/java/mods/railcraft/common/util/misc/WhiteBlackList.java
@@ -11,10 +11,13 @@
 package mods.railcraft.common.util.misc;
 
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
 
 /**
  * A combination of black list and white list.
+ *
+ * The whitelist has a higher priority than the blacklist.
  */
 public final class WhiteBlackList<T> {
 
@@ -27,7 +30,7 @@ public final class WhiteBlackList<T> {
     }
 
     public static <T> WhiteBlackList<T> create(Set<T> blacklist) {
-        return new WhiteBlackList<>(blacklist, Collections.emptySet(), false);
+        return new WhiteBlackList<>(blacklist, new HashSet<>(), false);
     }
 
     private WhiteBlackList(Set<T> blacklist, Set<T> whitelist, boolean whitelistEnabled) {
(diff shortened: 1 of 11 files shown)


src/main/java/mods/railcraft/common/worldgen/OreGeneratorFactory.java
```diff
@@ -213,7 +222,7 @@ public boolean isRichBiome(Biome biome) {
                 return level == WhiteBlackList.PermissionLevel.WHITELISTED;
 
             // FIXME this is wrong, it doesn't have to match all
```
CovertJaguar: Might want to remove the FIXME

src/main/java/mods/railcraft/common/util/misc/WhiteBlackList.java
```diff
@@ -58,16 +61,16 @@ public boolean isWhitelistEnabled() {
     }
 
     public boolean permits(T entry) {
-        return (!whitelistEnabled || whitelist.contains(entry)) && !blacklist.contains(entry);
+        return (whitelistEnabled && whitelist.contains(entry)) || !blacklist.contains(entry);
```
CovertJaguar: The blacklist should have priority according to the documentation in the cfg files.

src/main/java/mods/railcraft/common/worldgen/OreGeneratorFactory.java
```diff
@@ -213,7 +222,7 @@ public boolean isRichBiome(Biome biome) {
...
-            return BiomeDictionary.getTypes(biome).stream().allMatch(richBiomeTypes::permits);
+            return BiomeDictionary.getTypes(biome).stream().anyMatch(richBiomeTypes::permits);
```
CovertJaguar: This whole thing is no good. If any type is blacklisted, it should not be allowed. This check only checks that there is a type that is allowed which is also not blacklisted.

For example, you could blacklist Forest, and whitelist Mountains. The documentation indicates that a Biome of type "Mountainous Forests" would be blacklisted because it has Forest as a type. But the current implementation allows it because it has Mountain as a type and mountain isn't Blacklisted, ignoring the fact that Forest is blacklisted.

A proper solution must first check if the biome list intersects the blacklist, and if not, then check if the biome list intersects the whitelist.

CovertJaguar (changes requested):
Still not quite right.

src/main/java/mods/railcraft/common/worldgen/OreGeneratorFactory.java
liach: Fixed.

src/main/java/mods/railcraft/common/util/misc/WhiteBlackList.java
liach: Resolved.

src/main/java/mods/railcraft/common/worldgen/OreGeneratorFactory.java
liach: Deleted.

CovertJaguar (approved):

PR #1763, opened 2019-03-07
Revert gregtechified structure stone recipes

### LINKED ISSUE #1576: Quarried Stone recipe too weak
Crafting Quarried Stone to Quarried Block seems to reduce the available quantity by 1/4. It is a 2x2 (4) to 1 ratio. From the law of conservation of mass, it should produce 4 Quarried Block.

Using railcraft-12.0.0-alpha-9

liach:
**The Issue**
 - Close #1576
 - Decoration blocks should not be too tech-dependent.
 
**The Proposal**
 - Revert gregtechified structure stone recipes
 - Also update code to modern usages of `getStack(IVariantEnum)` instead of using raw meta. 
 
**Possible Side Effects**
 - Balance might differ, though this should not affect too much (structure module is relatively minor)
 
**Alternatives**
 - I do not want people to suffer for some simple decoration, hence this.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/aesthetics/brick/BrickTheme.java
+++ b/src/main/java/mods/railcraft/common/blocks/aesthetics/brick/BrickTheme.java
@@ -10,11 +10,12 @@
 package mods.railcraft.common.blocks.aesthetics.brick;
 
 import mods.railcraft.api.core.IVariantEnum;
-import mods.railcraft.api.crafting.Crafters;
 import mods.railcraft.common.blocks.IRailcraftBlock;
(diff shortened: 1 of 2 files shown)


CovertJaguar:
I may need to load this into a world and see the differences myself. I'm not really getting a feeling for it from looking at the code. Will deal with later.

liach:
i just changed stone recipes to the old ones before camo kicked in.

CovertJaguar:
Ah... k, I'm not sure I even know what he changed them to.

CovertJaguar:
I re-added the raw stone to cobble recipes, they seemed useful and actually make it easier to acquire cobble (no need to cook first). Kept all the other changes.

PR #1774, opened 2019-03-12
Moved all recipes to init

### LINKED ISSUE #1610: [1.12.2] Can't remove anything from Railcraft via Crafttweaker
I tried to remove recipes from Railcraft but it don't work. It seems that Railcraft recipes can't be removed.
Here is a simple script. -> https://pastebin.com/WuEMQhp2

Both commands did not work. recipes.remove and recipes.removeByRecipeName.

Forge: 2796
Crafttweaker: 4.1.13
Railcraft: 12.0.0 Beta 2

See also: https://github.com/CraftTweaker/CraftTweaker/issues/691

liach:
**The Issue**
 - Close #1610.
 - Describe the problem if an issue doesn't exist. Be as clear as possible.
 
**The Proposal**
 - This is to bring crafttweaker compatibility.
 - I inserted a log message and now confirm all recipes have been moved to init.
 
**Possible Side Effects**
 - Should be fine as mods already finish all registry things during preinit.
 
**Alternatives**
 - Might consider asking ct author for what else to do.

P.S. Forgot to open a pull for this change a few days ago. :sweat:

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/aesthetics/glass/BlockStrengthGlass.java
+++ b/src/main/java/mods/railcraft/common/blocks/aesthetics/glass/BlockStrengthGlass.java
@@ -90,8 +90,15 @@ public class BlockStrengthGlass extends BlockGlass implements IRailcraftBlock, C
     }
 
     @Override
-    public void finalizeDefinition() {
-        IRailcraftBlock.super.finalizeDefinition();
+    public void defineRecipes() {
+        for (EnumColor color : EnumColor.VALUES) {
+            CraftingPlugin.addShapedRecipe(getStack(8, color),
+                    "GGG",
+                    "GDG",
+                    "GGG",
+                    'G', RailcraftBlocks.GLASS.getWildcard(),
+                    'D', color.getDyeOreDictTag());
+        }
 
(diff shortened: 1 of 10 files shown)


src/main/java/mods/railcraft/common/blocks/aesthetics/glass/BlockStrengthGlass.java
```diff
@@ -90,8 +90,15 @@ public void initializeDefinition() {
     }
 
     @Override
-    public void finalizeDefinition() {
-        IRailcraftBlock.super.finalizeDefinition();
```
CovertJaguar, line 94: Removing this breaks the color handler.

CovertJaguar (changes requested):

src/main/java/mods/railcraft/common/blocks/aesthetics/glass/BlockStrengthGlass.java
liach: This is not removed; as the implementation for this method is removed, `finalizeDefinition` now inherits the default implementation from `IRailcraftBlock`.

CovertJaguar (approved):

PR #1775, opened 2019-03-12
Adds a proper api jar and api sources jar for mods to use

liach:
**The Issue**
Cannot find a proper railcraft api jar for gradle dependency. I encountered this issue during development of steves carts 2 for railcraft compatibility.
See https://github.com/TechReborn/StevesCarts/pull/179
 
**The Proposal**
Add api jar with srg names and api source jars with mcp names under the latest stable 32 for mc 1.12.2 for mods to use as a flatdir dependency.
 
**Possible Side Effects**
Other mods can still use the code from railcraft or just use rc api from github. old zip users should be able to extract source jar as well.
 
**Alternatives**
In long term, I suggest setting up a maven or using bintray for maven repos.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -325,14 +325,27 @@ task signDevJar(type: JarSigner, dependsOn: 'devJar') {
     outputFile = devJar.getArchivePath()
 }
 
-task apiZip(type: Zip) {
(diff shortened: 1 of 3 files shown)


liach:
Also improved code for multiblock a little bit.

Issue #1116, opened 2017-03-19
Switch motors not rendering


brisingraerowing:
I did a custom build from the repository to test out some of the new features (I had to disable the fingerprint check to get MC to load. I tried to figure out how to create a self-signed certificate, but got all befuddled while researching that), and quickly found that the switch motors render in the inventory, but not as a block. They still have a hitbox and they work perfectly fine, they just can't be seen. Are the block models for them not done yet? Or is there a bug?

Overall, though, the switches and junctions are quite cool. I am going to use them in a completely automated super factory thing (Crafting, Brewing, Smelting, Mining, etc) to route carts to various places. I always hated the Vanilla T-Switches, they are way too limited for any decent automation / rail system.

This is an awesome mod. Keep up the good work!

CovertJaguar:
Works here, they were broke at one point in the described manner, but I don't even know if I ever committed that version to the repo.

brisingraerowing:
I am guessing you didn't. I also noticed that switches only want to point towards the east, and can't be rotated to point any other direction. Same goes for Wye tracks. Is that by design? Seems to be rather limiting.

CovertJaguar:
No idea what you mean. You are supposed to just lay out the track how you want and use the Spike Maul on the track in the middle.

The version with the fixed rendering was only one I think I committed.

brisingraerowing:
Ah, user error then.

brisingraerowing:
I'll get a screenshot of what I am seeing (or not seeing) and a list of my mods up soon. Just have to wait for MC to load.

brisingraerowing:
Here is a [screenshot](https://1drv.ms/i/s!Ag_TFYG4M5mArj8rJKII-nafoTZy) and here is my [modlist](http://www.modlister.com/user/BrisingrAerowing).

The switch actuator works perfectly fine. It just is invisible.

brisingraerowing:
A quick test with just Railcraft and JEI confirms that this is a compatibility issue. I am going to go through my modlist and see what is broken.

brisingraerowing:
Chisel is behind the bug. I will report this to them.

tterrag1098:
@CovertJaguar Would you mind pointing me to the Block class for switches? I have an idea of what may cause this, but I need to read through the code.

tterrag1098:
I found the issue, you are relying on states passed into your model being non-extended, which is a bad assumption 

[...]

PR #1777, opened 2019-03-13
Fix rendering of item form of actuators

liach:
**The Issue**
Acturator item forms only had base but not the arrows rendering.
 
**The Proposal**
Moved actuator model loader to be controlled by the block type
Now we have many models instead of one; we do not use block state from world any more as that cannot carry any info for item form.
Removed useless inventory and normal variants. They are not used.
Dumped useless actuator item class and override methods in the enum class itself (avoid making an inner class)
 
**Possible Side Effects**
Shouldn't affect anything outside of actuators.
 
**Alternatives**
Not good ones that I am aware of.


liach:
This is the fixed screenshot.

src/main/java/mods/railcraft/client/render/models/resource/ActuatorModel.java
```diff
@@ -92,70 +108,92 @@
...
+        public CompositeModel(IBakedModel baseModel, IBakedModel redFlagModel, IBakedModel whiteFlagModel) {
+            this.baseModel = baseModel;
+            this.redFlagModel = redFlagModel;
+            this.whiteFlagModel = whiteFlagModel;
         }
 
         @Override
         public List<BakedQuad> getQuads(@Nullable IBlockState state, @Nullable EnumFacing side, long rand) {
-            if (state instanceof IExtendedBlockState)
-                state = ((IExtendedBlockState) state).getClean();
```
CovertJaguar, line 151: This is necessary, see #1116

CovertJaguar:
Could we also get a first person screenshot of holding one in hand?

And you mention "lots of models" but I don't see any additional JSONs added. Am I misunderstanding what you mean?

CovertJaguar (changes requested):

src/main/java/mods/railcraft/client/render/models/resource/ActuatorModel.java
liach: We no longer use the state here.

CovertJaguar (approved):

PR #1920, opened 2020-02-21
Fixed #1804 Charge networks fail to build properly

### LINKED ISSUE #1804: Charge Net Failing to Build
That title is horrible, but I can't come up with a better word.

I'm using electric track exclusively.  Frequently, but by no means dependably, I have had contiguous sections of track end up on different charge networks, typically with one or more "Size: 0" charge networks between them.

So, to reiterate via an example.  Right now, I have a length of electric track, around 600 blocks long (ish, there are corners).  There are no breaks / missing tracks.  If I use the Charge Meter on one segment, it reports a size of 331.  If I use the Charge Meter on a different segment, it reports a size of 357.  Between the two segments, there are two tracks that each report a size of 0.

Power applied to one segment will not (unsurprisingly) propagate to the other segment. 

Harvesting and replacing the two "size 0" tracks will join the two segments into one single net.

--

Steps to Reproduce:

Yeah, right.  I wish.  I've not found any dependable means to reliably recreate the disconnection.    I will say that when placing the track, I can be a bit haphazard about it, typically just running along, holding down the mouse button, then going back to fill in any tracks that didn't get placed.  I am not willing to promise that this behaviour causes, or even exacerbates, the issue described.

Once a network is created and stable, it will not subsequently break on its own.  It only happens when adding (and removing?) track. 

That said, the "size 0" tracks are not necessarily the track that has been recently added.

--

I'm not sure I've done an amazing job of describing what I'm seeing.  If I have been less than clear, I'll try to clarify any particular bits that need it.

--

No crash report.

Railcraft 12.0 for Minecraft 12.2
Forge 2811

ParapetCloud:
**The Issue**

 Charge networks fail to build properly. #1804 

**The Proposal**

I added a check to see if a node still needed to be added in AddNodeImpl
Sometimes when a node is in queue to be added, something will try to access the node. when this happens the node is added twice (once by the access, and then again by the queue) and the new node can end up pointing at a grid it is not a part of. I believe this is caused by the node being added to the grid twice.

I also added a check to grid.add to prevent invalid nodes from being copied to a new grid.
 
**Possible Side Effects**

it is possible that if a node incorrectly still existed in the location of the new node that it will not be updated, however the check I added is already used for screening in AddNode()
 
**Alternatives**

One alternative was to have a loop checking x number of nodes per tick to make sure it's pointing at the grid it's actually a part of. but this is costly in terms of time and isn't a true fix of the problem

Another alternative is to remove the node creation from ChargeNetwork.Access but I was unsure of what consequences this would cause from returning NULL_NODE for a position that will eventually have a node

one last alternative would be to remove the node from the queue when it is added in AddNodeImpl. but this once again, sounds costly for something that isn't going to happen that often.

### DIFF
--- a/src/main/java/mods/railcraft/common/util/charge/ChargeNetwork.java
+++ b/src/main/java/mods/railcraft/common/util/charge/ChargeNetwork.java
@@ -134,26 +134,28 @@ public class ChargeNetwork implements Charge.INetwork {
      * Add the node to the network and clean up any node that used to exist there
      */
     private void addNodeImpl(BlockPos pos, ChargeNode node) {
-        ChargeNode oldNode = nodes.put(pos.toImmutable(), node);
-
-        // update the battery in the save data tracker
-        if (node.chargeBattery.isPresent())
-            worldData.initBattery(node.chargeBattery.get());
-        else
-            worldData.removeBattery(pos);
-
-        // clean up any preexisting node
-        if (oldNode != null) {
-            oldNode.invalid = true;
-            if (oldNode.chargeGrid.isActive()) {
-                node.chargeGrid = oldNode.chargeGrid;
-                node.chargeGrid.add(node);
+        if(needsNode(pos, node.chargeSpec)){
+            ChargeNode oldNode = nodes.put(pos.toImmutable(), node);
+    
+            // update the battery in the save data tracker
+            if (node.chargeBattery.isPresent())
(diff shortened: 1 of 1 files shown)


liach:
imo just fast return if the need check fails would make the code much cleaner.

liach (approved):
I shall approve, but we need to wait for CovertJaguar.

Issue #1954, opened 2020-06-09
Server crash when player has a Thruster from Futurepack in their inventory


Maxwell-lt:
**Description of the Bug**
Server crash when a player has a Futurepack thruster in their inventory

**To Reproduce**
1. Use creative menu (or the crafting recipe) to add a Thruster block from Futurepack to your inventory
2. Server crashes with exception `java.lang.IllegalArgumentException: Cannot set property PropertyInteger{name=metadata, clazz=class java.lang.Integer, values=[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]} to 13 on block fp:triebwerk, it is not an allowed value`

**Expected behavior**
The server does not crash.

**Logs & Environment**
Minimum environment to reproduce:
- Minecraft 1.12.2
- Forge 14.23.5.2854
- Railcraft 12.0.0
- Futurepack 1.12.2-26.4.89

Logs:
 - Full log: https://gist.github.com/Maxwell-lt/28ae54f51a5c6697b0fa037d4a76f038#file-full-logs
 - Crash report (If available): https://gist.github.com/Maxwell-lt/28ae54f51a5c6697b0fa037d4a76f038#file-crash-log

**Additional context**
The issue appears to be caused because Thrusters render facing upward in the inventory, but cannot by placed in that orientation as a block. Because of this, the call to getStateFromMeta(stack.getItemDamage()) fails. A fix is to catch IllegalArgumentException and return the default value of Blocks.AIR.getDefaultState() instead.

PR #1955, opened 2020-06-09
Fix crash with Futurepack thrusters in a player's inventory

Maxwell-lt:
Futurepack thrusters render facing up in the inventory, but cannot be placed as a block in that orientation.
This patch catches the rare case that a call to `block.getStateFromMeta(stack.getItemDamage())` throws an exception.
Without this patch, a player with a thruster in their inventory will cause a server crash.

**The Issue**
This PR fixes issue #1954 
 
**The Proposal**
In the rare case that calling getStateFromMeta on an itemstack throws an exception, this will return the default value of Blocks.AIR.getDefaultState().
 
**Possible Side Effects**
This change has the effect of treating blocks that would otherwise crash the game as air blocks, which may cause undesired behavior if the Thrusters are used in Railcraft machines. However, the alternative is crashing the server.
 
**Alternatives**
This issue may be fixable from the side of Futurepack, but a similar issue raised against that project was rejected as the other mod's fault. Patching this crash on Railcraft's side was also a matter of a few lines, making this the faster route to a fix.

### DIFF
--- a/src/main/java/mods/railcraft/common/util/inventory/InvTools.java
+++ b/src/main/java/mods/railcraft/common/util/inventory/InvTools.java
@@ -485,7 +485,11 @@ public abstract class InvTools {
         Item item = stack.getItem();
         Block block = GameData.getBlockItemMap().inverse().get(item);
         //noinspection deprecation
-        return block == null ? Blocks.AIR.getDefaultState() : block.getStateFromMeta(stack.getItemDamage());
+        try {
+            return block == null ? Blocks.AIR.getDefaultState() : block.getStateFromMeta(stack.getItemDamage());
+        } catch (IllegalArgumentException e) {
+            return Blocks.AIR.getDefaultState();
+        }
     }
 
     public static @Nullable IBlockState getBlockStateFromStack(ItemStack stack, World world, BlockPos pos) {
--- a/src/main/java/mods/railcraft/common/util/inventory/InvTools.java
+++ b/src/main/java/mods/railcraft/common/util/inventory/InvTools.java
@@ -484,8 +484,8 @@ public abstract class InvTools {
             return Blocks.AIR.getDefaultState();
         Item item = stack.getItem();
         Block block = GameData.getBlockItemMap().inverse().get(item);
-        //noinspection deprecation
         try {
+            //noinspection deprecation
             return block == null ? Blocks.AIR.getDefaultState() : block.getStateFromMeta(stack.getItemDamage());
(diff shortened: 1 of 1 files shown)


src/main/java/mods/railcraft/common/util/inventory/InvTools.java
```diff
@@ -484,8 +484,12 @@ public static IBlockState getBlockStateFromStack(ItemStack stack) {
             return Blocks.AIR.getDefaultState();
         Item item = stack.getItem();
         Block block = GameData.getBlockItemMap().inverse().get(item);
-        //noinspection deprecation
-        return block == null ? Blocks.AIR.getDefaultState() : block.getStateFromMeta(stack.getItemDamage());
+        try {
+            //noinspection deprecation
+            return block == null ? Blocks.AIR.getDefaultState() : block.getStateFromMeta(stack.getItemDamage());
```
liach, line 489: Minecraft vanilla contract never said that `getStateFromMeta` should throw exception. How do we know if it throws illegal argument, unsupported operations, or whatever? This is not a "fix". Whoever throws the exception should fix it on their side by changing the implementation to call `return this.getDefaultState()` instead!

liach (changes requested):

liach:
Futurepack bug, not our fault at all. They break implementation contracts. You can "fix" this issue in one mod like railcraft but other mods will still break. This is not a solution at all. The solution is for future pack to replace `throw new IllegalArgumentException` with a `return this.getDefaultState()`.

src/main/java/mods/railcraft/common/util/inventory/InvTools.java
mcenderdragon: Hi. It is true getStateFromMeta should never throw an exception, but its also true that you can not assume item metadata = block state metdata. It is surprising only railcraft had this bug so far. Is it possible to use `getStateForPlacement` instead?

Castone22: @mcenderdragon It may be safe to assume this in the context this code seems to be used for.  It seems to be a hook that railcraft uses to determine whether the player should receive an advancement for having a given item in their inventory, which this should be a an adequate check for, especially if railcraft follows an internal contract where item meta = block meta.

Regardless this is something that has a high chance of blowing up on other mods since it does break the established interface for how a block should behave, it may just be luck that we haven't encountered it in anything else yet.

PR #1963, opened 2020-07-26
Rock crusher fix and some block interaction events.

Radviger:
**The Issue**
Rock crusher destroys items dropped on top when it's inventory is full.
Block interaction FakePlayer uses name `[railcraft]` which is no longer supported by many region protection mods/plugins due to uuids.
 
**The Proposal**
Fixed rock crusher bug.
Added missing block interaction events (eg. for firestone) and made them use their owner's profile (the player that placed block/cart/etc.)
 
**Possible Side Effects**
I've accidentaly removed mod signature checking in obfuscated env during testing. Because I've already commited it, it must be rolled back manually.

### DIFF
--- a/src/main/java/mods/railcraft/common/blocks/logic/TradeStationLogic.java
+++ b/src/main/java/mods/railcraft/common/blocks/logic/TradeStationLogic.java
@@ -10,6 +10,7 @@
 
 package mods.railcraft.common.blocks.logic;
 
+import com.mojang.authlib.GameProfile;
 import mods.railcraft.api.core.RailcraftFakePlayer;
 import mods.railcraft.common.gui.EnumGui;
 import mods.railcraft.common.plugins.forge.VillagerPlugin;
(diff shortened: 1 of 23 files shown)


CovertJaguar:
A lot of your changes make no sense to me.

CovertJaguar:
You didn't PR the API changes, and the Rock Crusher change should really have been a separate PR.

I cannot merge this in its current state.

CovertJaguar:
Technically this should be three PRs:

- The Rock Crusher change. Which is working as intended and any such change would be rejected.
- The FIX for the Fake Player ID. Which I don't see because its in the API repo, nor do I see any recommendation for how to fix it. I'm not familiar enough with Protection plugins.
- The additional protection checks.

CovertJaguar:
I have added my own version of the additional region protection checks you tried to add. The rest of it make no sense to me so I have closed this.

Radviger:
I see your point about separate PRs and I probably have to make them as you've suggested. So my question is: Is that actually an intended behaviour for rock crusher to just destroy any item stack that can't fit in?

Trinsdar:
imo it kinda makes sense since it's a rock crusher, you should be careful how you set it up. That's just my opinion though, not the official stance

Radviger:
I'm not saying it crushes them into nothing, I'm saying that any item you throw on top of it just pops out of existence if all the 9 processing slots are filled with items of other types.

Radviger:
IMO i'ts a BUG.

CovertJaguar:
I can see an argument for not destroying items it could crush, but that wasn't what the submitted code did. It made it not destroy anything.

Its supposed to destroy items it can't crush. This is the intended bit.

Issue #1388, opened 2018-04-03
Add back MJ-producing engines


bookerthegeek:
Edits by liach: This issue will, from now on, dedicate on tracking the progress of engines, instead of tracking in #1373. The Railcraft team has decided that when three steam engines are back, they will be producing microjoules. However, we are awaiting on BuildCraft's stablization.

Original issue:
Since BC has gone back to using MJ instead of RF, would it be possible to add an engine of some sort to output MJ as we will already have one for EU & FU/FE?

Thank you for taking the time to consider this request.

3TUSK:
#1232 ?

bookerthegeek:
Almost. The previous versions of BC used the COFH api for power, but no longer.

Also they do not use the FU/FE either to my knowledge.

liach:
this is likely only for 1.12

AlexIIL:
Just FYI the BC MJ API isn't finalised yet, so I don't recommend adding anything *yet*.

liach:
@AlexIIL Adding bc api as submodule is safe, right?

AlexIIL:
Sort of. I'd recommend using our maven instead though (see https://www.mod-buildcraft.com/pages/maven.html for details) although the api will change in the future, so you would need to ensure you only use safe bits (as far as I know railcraft only uses the trigger and action API, which should be basically final - please check before using anything else, as it might change if I don't think anybody's using it).

bookerthegeek:
As a small request in addition, would it be possible to make it a 2x3 multiblock similar to the EU turbine instead of a single block?

Generalcamo:
Working on it now. Going to reimplement the engines first, then we can see about an MJ to Charge multiblock.

liach:
FYI you should make a buildcraft module.

Generalcamo:
Just to clarify: I'm not adding support for MJ just yet. The API is too unstable for that. I will however work on the engines as the MJ API is worthless without them.

Also this is no confirmation on that multiblock. I myself am not too keen on converting pneumatic power directly to charge; at the very least it should be pretty lossy.

liach:
We are now sure that when the steam-powered engines are back, they will be producing MJ. Generalcamo is doing the most model design work.

tungstonminer:
[...]

PR #2000, opened 2021-01-21
Re-Implement MJ-Producing Engines

Su5eD:
## The Issue
See #1388 for details
 
## The Proposal
- Added back the Hobbyist's, Commercial, and Industrial Steam Engine
- They now produce MJ at the same rates at they did before the switch to RF. This includes the maximum amount of extractable energy, which is no longer 8 times the output, but is rather constant for each of the machines. 
- Rendering is still done using a TESR

Aside of these changes, the engines function the same way the did in 1.7.10.
Let's summarize the MJ rates:

|                         | Hobbyist's Steam Engine | Commercial Steam Engine | Industrial Steam Engine |
| :--------------: | :------------------------: | :--------------------------: | :-----------------------: |
|       Output      |                    2                 |                     4                    |                  8                  |
| Buffer Capacity |               10000             |                 20000                |               30000             |
|   Max Receive   |                 300               |                   600                  |               1200               |
|    Max Extract   |                  40                |                    80                   |                160                |
 
### Localization
A [PR](https://github.com/Railcraft/Railcraft-Localization/pull/354) is open in Railcraft-Localization with updated tooltips for the engines.

## Possible Side Effects
None that I know of
 
## Alternatives
I was thinking about keeping the current way of calculating the max energy extraction amount, but since the plan was to make the engines produce MJ again, I decided to use the same way they did before the switch to RF/FE.


src/main/java/mods/railcraft/common/blocks/single/TileEngine.java
```diff
@@ -410,4 +376,12 @@ public static EnergyStage fromOrdinal(int ordinal) {
         }
 
     }
+
+    @Optional.Interface(iface = "buildcraft.api.mj.IMjConnector", modid = "buildcraftlib")
```
aimiebell, line 380: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

Su5eD: I'm sorry, but your detection is wrong. What you've highlighted is an annotation, and following class name coventions, its name should be PascalCase, which already is. [Nm: Class names should start with an upper case letter (NM_CLASS_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-should-start-with-an-upper-case-letter-nm-class-naming-convention)

PR #2035, opened 2022-01-09
Tank cart fix

### LINKED ISSUE #7393: Manually filling tank cart with polyethylene show wrong amount of liquid
#### Which modpack version are you using?
2.1.0.0

#
#### If in multiplayer; On which server does this happen? 
private

#
#### What did you try to do, and what did you expect to happen?
1. Place empty tank cart
2. Right-click in it's tank with polyethylene  cell
3. Observe cart tank having 144L of  polyethylene 

#
#### What happened instead? (Attach screenshots if needed)
3. Observed cart tank having 13,158,600L of  polyethylene 

image
image

#
#### What do you suggest instead/what changes do you propose?
No clue what may cause it, maybe some cell output value configuration requires fixes? 

Looks like visual bug to me, liquid is not actually created from thin air, clicking on the tank with empty cell yields exactly one cell.

Also tested with Hydrogen Gas and it displays expected value. Also getting fluid back using cell works ok and removes all liquid.

YannickMG:
Fixes GTNewHorizons/GT-New-Horizons-Modpack#7393


YannickMG:
Wrong repo, apologies.

Issue #1411, opened 2018-05-18
[1.12.2] Missing recipes in JEI


bookerthegeek:
## Heyo
Here are some recipe related issues;
- Rock Crusher Recipes not showing up in JEI
- Rolling Machine recipes not Showing up in JEI
- Powered Rolling Machine crafts Small Charge Wire Spools for Free ([Image](https://i.imgur.com/gIaMXaa.png))

Working but not in JEI;
- Wooden Tie
- Creosote Wood Block
- Refined Firestone
- Metal Posts

Can not find recipes for the following;
- Force Track Emitter
- Trade Station
- Large Charge Wire Spool
- Any Electrode
- Charge Shunting Wire
- Track Parts
- Blank Tickets

liach:
machines recipes are temporarily removed from jei because i don't understand jei that well yet. a few working recipes do not show up because they are of custom types.

liach:
Now everything except JEI conpat is fixed.

xfl03:
The third problem may has been fixed in https://github.com/Railcraft/Railcraft/commit/101b6cde9c48d1ea717e624020ab99fa144694c8

liach:
Yep. Actually before your commit, it already stops giving free items (or stops giving any output) :trollface:

liach:
This should be totally fixed by #1487. I will test.

liach:
According to the picture in #1494, the jei integration should be complete now. Closing.

TimGoll:
Can confirm, the JEI integration seemed complete to me

Carnaxus:
A quick addendum: Adding Just Enough Resources at least partially reintroduces this bug; I couldn't get the rolling machine recipes for the charge wire coils to show up until I removed JER.

kappa-maintainer:
@Carnaxus Could you post your whole mod list? I have this problem too, but can't fix it by removing JER

Carnaxus:
> @Carnaxus Could you post your whole mod list? I have this problem too, but can't fix it by removing JER

@kappa-maintainer

Advanced Machines 61.0.0 (IC2 addon)
Advanced Machines Patcher 1.0
Advanced Solar Panels 4.3.0 (IC2 addon)
AdvSolarPatch 1.2.1
BetterFPS 1.4.8
Buildcraft 7.99.24.8
Chameleon 4.1.3
ChickenChunks 2.4.2.74
Chisel 1.0.2.45
CodeChickenLib 3.2.3.358
[...]

PR #2049, opened 2022-06-11
Fix #2042

### LINKED ISSUE #2042: JEI not showing Railcraft recipes with some mods
A quick addendum: Adding Just Enough Resources at least partially reintroduces this bug; I couldn't get the rolling machine recipes for the charge wire coils to show up until I removed JER.

_Originally posted by @Carnaxus in https://github.com/Railcraft/Railcraft/issues/1411#issuecomment-1075796047_

kappa-maintainer:
**The Issue**
#2042 did not get fixed by removing onRuntimeAvailable
 
**The Proposal**
This is a change prompt by IDE when I was removing onRuntimeAvailable. Tbh, I have no idea why it's essential

Should look at https://github.com/Railcraft/Railcraft/issues/1411#issuecomment-1152733715 for a triggering mods combination
 
**Possible Side Effects**
Probably no side effects.
 
**Alternatives**
No unless I want to dive into multi-mod JEI interaction

### DIFF
--- a/src/main/java/mods/railcraft/common/plugins/jei/RailcraftJEIPlugin.java
+++ b/src/main/java/mods/railcraft/common/plugins/jei/RailcraftJEIPlugin.java
@@ -64,7 +64,7 @@ public class RailcraftJEIPlugin implements IModPlugin {
     public static final String ROCK_CRUSHER = "railcraft.rock.crusher";
     public static final String COKE = "railcraft.coke";
 
-    private final List<IRecipeCategory<?>> categories = new ArrayList<>();
+
 
     @Override
     public void registerIngredients(IModIngredientRegistration registry) {
@@ -161,6 +161,8 @@ public class RailcraftJEIPlugin implements IModPlugin {
 
     @Override
     public void registerCategories(IRecipeCategoryRegistration registry) {
+        List<IRecipeCategory<?>> categories = new ArrayList<>();
+
         IJeiHelpers jeiHelpers = registry.getJeiHelpers();
         IGuiHelper guiHelper = jeiHelpers.getGuiHelper();
 


Carnaxus:
New note for ya kappa: I just threw together a Pixelmon-based instance yesterday that doesn't include Railcraft and am having issues with Pixelmon recipes not showing up unless, once again, I remove JER.  Looks like this may actually be a problem on JER's end, not Railcraft, Pixelmon, or other mods.

kappa-maintainer:
I will call it a conflict, since there are more mods working well with JER

kappa-maintainer:
So I test through all mods again and found the cause of it
It was the new [Witchery](https://witchery.msrandom.net/home/downloads) mod, when install it with Railcraft, all jei categories will disappear
This PR somehow fix it, I am struggling to figure out why

Carnaxus:
Ok that sounds a little more severe than what I was getting!  I don't use Witchery though...

The Pixelmon instance I mentioned has very few mods in it.  I'll list them out tomorrow when I'm back on my computer (on phone doing one last email check before sleep).

kappa-maintainer:
I noticed that the log said this after applied the fix, not sure what it means
Found a missing id from the world railcraft:emblem

CovertJaguar:
I'm not sure, maybe because everything is lazily instantiated.

MsRandom:
The reason it broke with Witchery: Resurrected(and the reason this fixes it), is that Witchery: Resurrected loads its recipes on world load and reload, not at startup, so we need to manually reload JEI after startup. Railcraft was only expecting JEI to load once, and the list was never cleared so it'd not work as expected, meanwhile when it's local it'll be a new list everytime.

Carnaxus:
One of the few downsides to having so many popular mods all with different authors, I suppose.

Issue #1094, opened 2017-03-05
Locomotive Types


Generalcamo:
# Primitive:
## Furnace Cart:
Basic engine running on the convection of heat and motion (read: magical handwaves) to operate the cart. Difficult to automate, but can handle early game shunting
## Accepts:
- Coal
- Charcoal

# Steam:
## Requirements:
-  Water
-  Fuel

## Low-Pressure Steam Locomotive:
Basic engine running on the heating of water to create steam which creates the power to operate the engine. Reliable design good for shunting and medium-range builds.
## High-Pressure Steam Locomotive:
 A more complicated engine running on the heating of water to create steam, which is further heated in a superheater to make superheated steam. Is more efficient, but requires a massive amount of fuel to start up and is dangerous if there is a failure. Good for long-range builds.

### Oil Burning:
Grants ability to use certain types of liquid fuel instead of solid fuels by converting the firebox. Cannot handle gasses or upward flowing liquids however as they would simply exit through the Chimney.

## Fuel Types:
- Anything in a furnace
- w/ Oil Burning upgrade: Lava, Creosote, Liquefacted Coal, anything that isn't a literal gas (marked as such in the fluid dictionary) or floats upward (marked as such in the fluid dictionary)

# Electric
## Requirements:
- Electric Track
- Railcraft Charge capability

## Electric Locomotive:
Engine that draws on outside power to operate. Highly efficient and reliable, but requires additional infrastructure in order to work well.
## Fuel Types:
- Railcraft Charge

# Diesel
## Requirements:
- Chem Lab 
## Diesel-Electric:
Engine that contains an onboard alternator to generate power for the engine from hydrocarbons. More efficient than an oil burner and reliable, but is expensive to operate due to the need to use processed hydrocarbons rather than any old type of fuel.
## Fuel Types:
- Diesel
- Ethanol
- Biodiesel
- other processed fuels

# Magic: 
## Essentia:
Engine that uses literal magic to power itself from Thaumcraft essentia.
## Fuel Types:
- Thaumcraft Essentia

dj3520:
My thoughts.

Would the oil burning upgrade be for Primitive and Steam or only steam?
Could the Magic locomotive also take Botania mana? I believe there's already a mana cart.
Perhaps an upgrade for the Diesel locomotive could be a Hybrid upgrade, where it also charges/discharges from an energy cart (charges when idling and discharges when more power is needed)

[...]

Issue #733, opened 2015-11-23
Diesel Locomotive?


excelcrafter:
Maybe new locomotive which can use diesel, fuel ( buildcraft ) biodiesel ( immersive engineering )
methanol, ethanol, or some other liquid fuel?

MFE cart is just too heavy to pull with electric loco to use it like mobile battery.

I was thinking about to use tank cart filled with steam as "fuel source" for steam loco, didnt try yet :D but I will try asap, haha

Forecaster:
Duplicate of #651

Issue #651, opened 2015-07-07
[Suggestion] Diesel fuel system?


tebruuls:
It would be cool if there was a diesel fuel system with engines, multiblocks to make it, and a new locomotive.

Forecaster:
This is already being worked on.

ElectroBot:
5 year bump

PR #2071, opened 2022-08-19
Add Diesel locomotive

barty32:
**The Issue**
There were mutiple requests to add a new Diesel locomotive (#1094, #2063, #651, #733), so i tried to implement it. It uses newly added fluid Diesel, but currently it is not possible to craft it (we need to make a system to produce it or use it from other mods).

**The Proposal**
Add a Diesel locomotive
 
**Possible Side Effects**
I don't know about any side effects.
Currently known issues / to-do list:
1. when switched from shutdown mode directly to running, the locomotive stays off
2. make engine + whistle sounds
3. make a way to craft diesel
4. add fuel compatibility from other mods
5. make a crafting recipe
6. translate newly added items into other languages
7. implement a temperature system (engine pre-heating, overheating?)

**Alternatives**

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,7 @@
 # This is required to provide enough memory for the Minecraft decompilation process.
 org.gradle.jvmargs=-Xmx3G
 org.gradle.daemon=false
-version=12.1.0-beta-7
+version=12.1.0-beta-8
 version_forestry=[5.8.0.256,)
 version_industrialcraft=[2.8.67-ex112,)
 version_buildcraft=[7.99.24.7,)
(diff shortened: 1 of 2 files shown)


Romz24:
Looks interesting, how can I test this?

UPD: I compiled from your sources, tested. Everything seems to work fine, thanks for trying to make the mod better.

Romz24:
Why did you decide to close it?

barty32:
Well this wasn't intentional 😀
I'm working on it now and as I moved those commits to a new branch it just broke this PR. I'll probably create a new one once it is done because there doesn't seem to be an option to reopen this.

Forecaster:
It can't be reopened because there are no commits on the branch it seems

Issue #2040, opened 2022-06-01
Steam Turbine can't form when IC2 Classic installed


kappa-maintainer:
**Description of the Bug**
RC will crash the game when install with IC2C

**To Reproduce**

1. Install both mods and launch the game
2. Build a Steam Turbine, log should begin spamming now
3. The multiblock will not assemble

**Expected behavior**
The multiblock should assemble

**Logs & Environment**
https://gist.github.com/kappa-maintainer/75a967522ee57ba4d384a570de86f355

**Additional context**
IC2C's API will manually trigger a block update when tileentities were added to the energy network (ic2.core.energy.EnergyNetLocal#addMetaTileEntity), which triggers StructureLogic.onBlockChange and eventually change the whole structure state to UNTEST when forming

CovertJaguar:
Probably related to:
https://github.com/Railcraft/Railcraft/commit/2dfec3e30ec230a12e7fc6190b62f3f6391b5c26

kappa-maintainer:
This may be fixed by doing an extra check on block update handling

PR #2086, opened 2022-10-27
Attempted Fix on Railcraft IC2Classic Support

Speiger:
IC2Classic Compat is not working. #2040
This is most likely due to a problem in how the API is used.

List of changes:
onStructureChanged changes: IsMaster is if the master is being called, this one is either always true or always false, but has no indication if the structure is valid or not. Meaning we only care about the master if we do changes, but the "isComplete" (is formed) is the indicator if the EU Stuff should be applied or not.

changes to add/drop: Basically trust the internal state, if that one is not trustworthy then you have other problems... Also just drop duplicated calls.

If a structure can be reformed while still being valid then we have different issues again.

A few things I couldn't figure out:
The master is expected to be "index = 0" in the "getSubTiles" list. As per IC2Exp specifications.
If revalidation attempts are send through the event?

Note that this patch may or may not work, I am throwing in a blind shot.
If it works great, otherwise just close this.


Speiger:
Confirmed by @xJon that the patch works.

